PR #8, opened 2018-05-13
Removes galaxy-simple

pcm32:
Removes the long deprecated galaxy-simple from the helm charts code part (it will remain in the io-pages).


pcm32:
Thanks for reviewing!

PR #9, opened 2018-05-13
Adds travis support.

pcm32:
Adds travis initially for helm linting, possibly adding chart packaging and publishing in the future.


pcm32:
Yes, that is where I'm aiming! Thanks!

afgane:
Should there be a README badge to go with this?

Issue #10, opened 2018-05-19
Galaxy-stable chart doesn't start up with default settings.


nuwang:
How to recreate:
I checked out the develop branch and tested against minikube. Did not make any changes - simply ran a `helm install .`

The galaxy-stable chart installs without a hitch. However, when attempting to access the galaxy service via the NodePort, it simply returned a 404. The galaxy container's logs revealed the error:
`
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) unable to open database file
`
It appears that the GALAXY_CONFIG_DATABASE_CONNECTION is not being overridden as expected.

I then exec'd into the container and manually ran the following:
```
export GALAXY_CONFIG_DATABASE_CONNECTION=postgresql://galaxy:change-me@postgresql-for-galaxy/galaxydb?client_encoding=utf8
/usr/bin/startup
```
Still the same error.

Finally, I edited galaxy-central/config/galaxy.ini, and set the database connection manually there. It then proceeded to startup and connected to the postgres database, instead of sqlite. However, it ran all the migrations from scratch, indicating that it's not a database with migrations included.

Finally, it failed with the following:
```
Traceback (most recent call last):
  File "/export/galaxy-central/lib/galaxy/webapps/galaxy/buildapp.py", line 58, in paste_app_factory
    app = galaxy.app.UniverseApplication( global_conf=global_conf, **kwargs )
  File "/export/galaxy-central/lib/galaxy/app.py", line 185, in __init__
    self.job_manager = manager.JobManager( self )
  File "/export/galaxy-central/lib/galaxy/jobs/manager.py", line 23, in __init__
    self.job_handler = handler.JobHandler( app )
  File "/export/galaxy-central/lib/galaxy/jobs/handler.py", line 33, in __init__
    self.dispatcher = DefaultJobDispatcher( app )
  File "/export/galaxy-central/lib/galaxy/jobs/handler.py", line 764, in __init__
    self.job_runners = self.app.job_config.get_job_runner_plugins( self.app.config.server_name )
  File "/export/galaxy-central/lib/galaxy/jobs/__init__.py", line 632, in get_job_runner_plugins
    rval[id] = runner_class( self.app, runner[ 'workers' ], **runner.get( 'kwds', {} ) )
  File "/export/galaxy-central/lib/galaxy/jobs/runners/kubernetes.py", line 44, in __init__
    assert KubeConfig is not None, K8S_IMPORT_MESSAGE
[...]

PR #11, opened 2018-05-25
Example configs for 18.01 minikube and container building script

pcm32:
- Example configs tested for lifting a vanilla Galaxy setup (18.01) with helm. 
- Container building script to produce community images containers with the correct settings, configurable to look into different galaxy repos and ansible-galaxy-extras.

Will add documentation on how to use this on the readme.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "docker-galaxy-stable"]
+	path = docker-galaxy-stable
+	url = https://github.com/bgruening/docker-galaxy-stable
--- a/README-galaxy-stable.md
+++ b/README-galaxy-stable.md
@@ -1,7 +1,18 @@
+# TL;DR
+
+To run a vanilla docker-galaxy-stable setup on Kubernetes using this chart on minikube:
+
+```
+helm install -f example_configs/galaxy-stable-18.01-minikube.yaml galaxy-helm-repo/galaxy-stable
+```
+
+This assumes that you at least followed the [Helm for the first time](README.md#galaxy-helm-charts) part and that you have checked out this repo (where the config file is available). Please note that this setup won't send jobs to the Kubernetes cluster, as it is not configured like that on the stock docker-galaxy-stable compose galaxy images. This feature requires changes in the `config/job_conf.xml`, which would mean that you need your own `galaxy-init` container flavour (like the [one built for PhenoMeNal](https://github.com/phnmnl/container-galaxy-k8s-runtime/blob/develop/Dockerfile_init) for instance).
+
+
+
 # Using docker-galaxy-stable compose images
 
-Using the docker-stable compose images with Kubernetes on this setup requires building those images with certain options being passed to the images, as done [here](https://github.com/phnmnl/container-galaxy-k8s-runtime/blob/develop/simplified_galaxy_stable_container_creation.sh). Please note that you will want to have your own version of `Docker_init` (like the one [here](https://github.com/phnmnl/container-galaxy-k8s-runtime/blob/develop/Dockerfile_init)). where you can setup your flavouring, tools and other settings to personalize the Galaxy instance that you will be deploying.
-
+Using the docker-stable compose images with Kubernetes on this setup requires building those images with certain options being passed to the images, as done [here](https://github.com/phnmnl/container-galaxy-k8s-runtime/blob/develop/simplified_galaxy_stable_container_creation.sh). Please note that you will want to have your own version of `Docker_init` (like the one [here](https://github.com/phnmnl/container-galaxy-k8s-runtime/blob/develop/Dockerfile_init)), where you can setup your flavouring, tools and other settings to personalize the Galaxy instance that you will be deploying.
 
 ## Variables
 
--- a/README-galaxy-stable.md
+++ b/README-galaxy-stable.md
@@ -1,6 +1,6 @@
 # TL;DR
 
-To run a vanilla docker-galaxy-stable setup on Kubernetes using this chart on minikube:
+To run a vanilla docker-galaxy-stable setup on Kubernetes using this chart on **minikube**:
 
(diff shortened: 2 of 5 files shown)


pcm32:
This addresses #10 .

bgruening:
That is super cool!
Have you any idea how we can test this with some public CI system? Is there any CI system used to test k8s setups or that can run VMs like minicube?

afgane:
We are pretty close to be being able to setup a k8s instance on demand from a vanilla Ubuntu machine so perhaps that can be used as the platform? It would also help make sure that setup is working as expected.

There's no documentation yet, but we use this playbook to bootstrap it all: https://github.com/CloudVE/ansible-cloudman2

pcm32:
For testing, is possible to install minikube directly on a Travis VM these days.

pcm32:
https://blog.travis-ci.com/2017-10-26-running-kubernetes-on-travis-ci-with-minikube

I have been meaning to put time on this for a while.

bgruening:
Awesome!!!!

nuwang:
Closes: https://github.com/galaxyproject/galaxy-kubernetes/issues/10

pcm32:
Thanks for trying it @nuwang. Current difficulty is that you need to add a modified version of job_config.xml (as we do in the PhenoMeNal init container) where the runner is activated. I don't think that this can be done by just environment variable setting (as you need to actually add chunks of XML), so it would probably require a k8s specific job_config.xml that we could add into a k8s specific init container. That would be a first step.

Then, for truly running in anger in Kubernetes, you need to setup resource usage for tools, otherwise you can meltdown your cluster. We do this manually in PhenoMeNal (see [here](https://github.com/phnmnl/phenomenal-h2020/wiki/Setting-up-Galaxy-wrappers-on-PhenoMeNal-Galaxy-Container#tool-cpu-and-memory-usage-requests-and-limits)), but we might need a way that scales better for general Galaxy images (or just setting a default, but then demanding tools will suffer/fail).

pcm32:
I wonder why you had to create your folder on minikube... I try it all the time with different prefixes and it always creates the folders on its own...

nuwang:
I'm not really sure - I'm running minikube v0.27.0 on a Mac. The PVC was not being satisfied, so I manually created the folders and it was happy.

Enis and I were just thinking about the minikube setup. Is it ok if we switch to the dynamic provisioner instead of using a manually defined PV? That way, we won't need a special config for minikube. Also, is there a specific version of Kubernetes that you want to support? If the latest is ok, we were thinking that that 1.6+ or so would be good, since a lot of stuff like storage classes are out of beta since that version. In our case, we were planning on 1.10.0+

pcm32:
But I don't think I'm supporting things earlier than 1.6 anymore in the helm charts.

pcm32:
The error you describe seems to fit with https://github.com/kubernetes/minikube/issues/2256 which apparently was fixed on 0.25 (I'm on 0.25.2 I think, on macOS), maybe there was a regression on 0.27.0??

pcm32:
I would suggest then that we make the minikube case the default, once advanced users try this locally, they might be more prone to spend money on a cloud provider to use it directly, will be willing to use more intricate settings and/or integrate it into their deployment method/cloud setup. For less advanced users, I don't even expect them to try direct deployments (but wrapper solutions like cloudman, phenomenal, etc). 

In my development experience with Galaxy-Minikube, is more useful to have the persistence storage working as you don't need to be uploading data each time to run tools or other configs, but it really doesn't matter I think, in the end you always end up using some additional setting that deviates from the default (like the init container that you will use for instance).

Should I make then the minikube setup the default? It would be nice to have a target Galaxy image to try this (so that tools work). Maybe an image with the tools used in usegalaxy.org? @bgruening will such a set of tools have all containers through bioconda/mulled? The only requirement would be that for all the tools that we add, automatic mapping to containers work. Where can I find the tool setup definition for those instances?

pcm32:
> @bgruening Was there a specific reason for having galaxy-central in /export? If we leave galaxy-central within the container, and not in /export, would that make upgrades easier? (i.e. It looks like swapping containers is not enough, we also have to delete the galaxy-central folder manually from /export for a galaxy update to take effect - is that correct?)

I have been arguing the same with @bgruening. I think that we need to move out of the Galaxy web - init duality, and have people override tools and other settings through an additional container starting from the base one and we avoid all the time the /export copy process which can add some 3 to 4 minutes to deployment based on the writing speed of the shared file system.

What I do for the updates re /export is that I added functionality to force the copy of certain elements to /export on startup of the container (you specify these folders in the helm chart), but I agree with @nuwang that certain things should live in the container and only be linked from /export, to avoid this.

pcm32:
I would suggest that the main separation of Galaxy containers in compose setup shouldn't be web - init, but web handler and jobs handlers, to truly scale on the amount of jobs that you can handle. This is probably not the best place to discuss that though.

nuwang:
@pcm32 Just to make sure I understood what you meant by: "have people override tools and other settings through an additional container starting from the base one". I take it you mean that we build the base galaxy container, and then extend that Dockerfile to create a new container with the desired tools installed? If so, that sounds good to me.

I'm still not sure how the tool install database will be handled here, unless the database is sqlite. Is that what you're thinking?

bgruening:
Hi,

the init-container needs to go away at some point, but there where technical limitation at the time we created them and there are still a few open Q. Happy to discuss this with all of you, whenever you have time.
One problem is, iff we want to move to handler-only containers we need to have a shared codebase, or we duplicate code across containers.

> I'm still not sure how the tool install database will be handled here, unless the database is sqlite. Is that what you're thinking?

Is the use-case the same as for the monolithic container? installing at first run will not work? If not the tool will be installed in the init-container and then moved into /export. At this time its fixed and you can not merge multiple flavors into one. But this was not possible before to my knowledge.

PR #14, opened 2018-06-02
Shifted to official postgres chart

nuwang:


nuwang:
@pcm32 Sure, I can roll back changes to the galaxy chart. However, it should be functionality equivalent, since it uses the same images as before, and only the kubernetes definitions are reused from the official postgres chart.

With regard to the PVC, yes, the "existingClaim" property allows for that: https://github.com/galaxyproject/galaxy-kubernetes/pull/14/files#diff-53aede5b82e957467958e748a2d5ffa2R92
I've tested this and everything appears to be in order.

pcm32:
Yes, they should be functionally equivalent, but I intend to completely move out of this chart (galaxy) once both phenomenal portal based deployment and development environment are based on galaxy-stable. So I wouldn't want anymore than bug fixes on galaxy chart to avoid losing any valuable development time from whoever is investing it.

I see that you started a new PR #17 which centers on galaxy-stable, should we close this PR to avoid confusion then?

Thanks!

nuwang:
@pcm32 Ok, thanks for clarifying. Closing this in favour of PR 17.

PR #18, opened 2018-06-04
Update chart conventions

nuwang:
Updates chart naming and definitions to more closely follow charts in official helm repo.

Migration guide to v2.0.0
-------------------------
1. postgres_for_galaxy section should be renamed to postgresql and subkeys changed as follows:
     db_password -> postgresPassword
     postgres_image -> image
    postgres_image_tag -> imageTag
    postgres_pvc -> persistence.existingClaim
    subPath -> persistence.subPath

2. The galaxy section has been removed and all sub keys promoted to top-level. This follows conventions from the official chart repo and also makes the chart  more composable. (e.g. galaxy.property, instead of galaxy.galaxy.property)
   e.g. galaxy.admin.email -> admin.email

   Some additional notable changes include:
   galaxy.brand -> galaxy_conf.brand
   galaxy.backend.postgres -> postgresql.enabled
   galaxy.smtp.server -> galaxy_conf.smtp_server
   galaxy.smtp.username -> galaxy_conf.smtp_username
   galaxy.smtp. password -> galaxy_conf.smtp_password
   galaxy.smtp.email_from -> galaxy_conf.email_from
   galaxy.smtp.ssl -> galaxy_conf.smtp_ssl
   galaxy.ftp.url -> galaxy_conf.ftp_upload_site
   galaxy.instance_resource_url -> galaxy_conf.url

    galaxy.internal_port -> service.port
    galaxy.node_port_exposed -> service.nodePortExposed

    galaxy.create_pvc -> persistence.enabled
    galaxy.pvc.name -> persistence.name
    galaxy.pvc.capacity -> persistence.size

3. Additional changes include
    ingress.hostname -> ingress.hosts array
    use_proftpd -> proftpd.enabled
    use_condor -> condor.enabled
    pv_minikube -> persistence.minikube.enabled
    pv.minikube.hostPath -> persistence.minikube.hostPath
    external_ingress_controller -> ingress.self_managed

4. galaxy config handling for galaxy.yml has been standardized. Arbitrary keys for galaxy.yml can be listed under the galaxy_conf section. These are eventually stored in a config map named -galaxy-conf.

5. Additional container settings can be included in galaxy_container_cfgmap.


index.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
```
pcm32: These files (index and the packaged tar.gz) don't go on the same branch as the code, please remove them from this branch and PR. Thanks!

afgane: Why not? Particularly during the current rapid development cycle it gives one point of reference vs. having to maintain and fish for a chart repo?

pcm32: For development, you would rather use `helm install -f config.yaml ./galaxy-stable` locally, no need for packaging while developing. We put the packaged files and index in the github-io branch.

pcm32: That is the convention followed when hosting helm on github...

pcm32: This branch should be only for the chart logic, not for storing its packaged version. Thanks!

nuwang:
@pcm32 I've added a migration guide which we'll keep updated as we go along.

index.yaml
afgane: OK, I had forgotten about the `gh-pages` branch. I'm using this primarily from Rancher, which requires a git repo. Instead of polluting that branch with repeated commits of the dev package then, I've just created a separate repo for the dev packages: https://github.com/CloudVE/helm-charts

pcm32:
Regarding moving more variables to config maps, I have been reluctant in the past as they don't tend to play well with helm charts rolling upgrades, but haven't retried this recently. So, if you change a value of a bar that goes in a configmap instead of the deployment object in an upgrade process, the deployment object doesn't get changed and its pods don't pick up the variable change. I would suggest that you test that, as it might mean that you want to move some variables off the configmap.

nuwang:
@pcm32 Thanks for raising that. This commit addresses that issue: https://github.com/galaxyproject/galaxy-kubernetes/pull/18/commits/d8fc53214225e5e78d7aa3e333f39eba4137c82b (based on: https://github.com/kubernetes/helm/blob/master/docs/charts_tips_and_tricks.md#automatically-roll-deployments-when-configmaps-or-secrets-change).

pcm32:
Another pending thing that I had re configmaps/secrets is that when doing a `helm delete <deployment-name>` the secrets and configmaps where not being deleted (and had to be manually deleted to do a subsequent install, mostly useful for development). At some point I remember reading that it was supposed to be intentional functionality on helm, but I wonder if that has changed. Might be good to address it in this major re-haul if possible.

nuwang:
WRT to the config maps/secrets, I think they were being left behind because some some of the metadata labels were missing. They seems to get cleaned up ok now.

nuwang:
The docs suggest that the default is an empty string, which simply mounts the folder without creating it: https://kubernetes.io/docs/concepts/storage/volumes/#hostpath
So perhaps it's tied to the version of minikube but having to specify the type seems like the new default behaviour.

pcm32:
I have tested this and make it to work with 18.05, completely independent of PhenoMeNal using this setup https://github.com/ebi-gene-expression-group/container-galaxy-sc-tertiary/blob/develop/helm-configs/tertiary-portals-galaxy-18.05-minikube.yaml

I'm happy to merge this as 2.0. I suggest that we merge the postgres PR and then this one. Will merge the postgres first, changing version there to 2.0.

pcm32:
Thanks @nuwang @afgane @almahmoud for such magnificent improvement to the helm chart!

PR #20, opened 2018-06-13
log to k8s

ilveroluca:
This PR:

1. lets stderr and stdout from `run_galaxy_config.sh` be captured by kubernetes, rather than sending only stdout to the file `config_db.log` and letting stderr go to k8s;
2. it removes the conditional creation of the sqlite directory since run_galaxy_config.sh already does that;
3. it sets `rbac_needed` to true by default.

### DIFF
--- a/galaxy/values.yaml
+++ b/galaxy/values.yaml
@@ -79,7 +79,7 @@ galaxy_backend_postgres: false
 
 legacy:
    pre_k8s_16: false
-rbac_needed: false
+rbac_needed: true
 
 k8s_supp_groups: ""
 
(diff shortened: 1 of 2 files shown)


galaxy/values.yaml
```diff
@@ -79,7 +79,7 @@ galaxy_backend_postgres: false
 
 legacy:
    pre_k8s_16: false
-rbac_needed: false
```
pcm32: My concerns on this one is that minikube with Kubernetes below 1.10 seems to have rbac disabled by default, where as minikube with k8s >= 1.10 has it enabled. Trying to start this on minikube k8s < 1.10 with `rbac_needed` will produce an error I think. Since k8s 1.9 is still widely in use, I would leave the default behaviour as it is.

ilveroluca: That sounds reasonable.  I'll change that.  Let's remember to document the fact that `rbac_needed` must be set to true for more recent versions of minikube.

PR #21, opened 2018-06-19
Creation of PV directory if needed for minikube

pcm32:
@ilveroluca this will sort out the issue you started to see regarding directory creation, courtesy of @nuwang.


pcm32:
Should we bump minor? given that this change might leave behind some k8s versions?

ilveroluca:
I think a minor version bump is warranted.  After all, it fixes a bug.

Issue #29, opened 2018-09-18
[2.0] welcome.html is not being served correctly


pcm32:
On the new helm chart setup, content is being served on `http://<host:port>/galaxy/` prefix. In this setting, adding a welcome.html page as describe in docker-galaxy-stable (as done [here](https://github.com/ebi-gene-expression-group/container-galaxy-sc-tertiary/blob/9a697471ff88634f5dfb5a8b87d188c1f458e913/Dockerfile_init#L24)) produces that Galaxy points the browser to find the html in `http://<host:port>/galaxy/web/welcome.html`, however nginx seems to be serving the page in `http://<host:port>/web/welcome.html`.

pcm32:
I think that this is produced because the nginx in the container is configured to use directly `/` where as UWSGI is configured to use `/galaxy`. The problem, seems to be, is that nginx configuration happens at container build time, whereas UWSGI setup happens at runtime through environment variables.

Any opinions @bgruening ??

pcm32:
Ok, I think I have sorted this out by adding uwsgi variables support on the helm chart. Will PR soon after finishing tests.

PR #30, opened 2018-09-18
Adds UWSGI env variable support and sets default to / instead of /galaxy

pcm32:
This PR:

- Adds the ability to set UWSGI_ environment variables for the configuration of UWSGI.
- Sets default case to serve galaxy at `/` instead of `/galaxy`. This default is chosen as the docker-galaxy-stable compose images seem to be configuring NGINX by default for this setup (and this is unfortunately done at container build time). Using `/galaxy`on UWSGI when NGINX is not set for this produces all set of weird HTML behaviour like #29 or broken http pulled file uploads, among others.

### DIFF
--- a/galaxy-stable/Chart.yaml
+++ b/galaxy-stable/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: "1.0"
 description: Galaxy is an open, web-based platform for accessible, reproducible, and transparent computational biomedical research.
 name: galaxy-stable
-version: 2.0.0
+version: 2.0.1
--- a/galaxy-stable/templates/galaxy_container_cfgmap.yaml
+++ b/galaxy-stable/templates/galaxy_container_cfgmap.yaml
@@ -38,6 +38,12 @@ data:
   "PROXY_PREFIX": "{{ .Values.ingress.path }}"
   {{- end }}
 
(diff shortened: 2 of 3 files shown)


pcm32:
I wonder if it would be relevant to bump the version of the chart to 2.0.1 or something, as the default is changed (although in fairness, previous default doesn't work well in my opinion, unless that you built your container differently).

pcm32:
I'm going to merge this since I haven't got a review quite yet.

pcm32:
Agreed! According to Bjeorn, nginx can be configured on runtime, but I didn't find the way. My understanding from documentation is that root deployment is the default, so I think it is sensible that the chart reflects that (while being able of course to do otherwise).

bgruening:
@pcm32 this here? https://github.com/bgruening/docker-galaxy-stable#configuring-galaxys-behind-a-proxy--toc

pcm32:
Yes, but that variable is set in the chart, but it doesn't work well because it only seems to affect the uswgi settings but not the nginx settings (see #29).

pcm32:
Sorry, forgot to paste https://github.com/galaxyproject/galaxy-kubernetes/blob/8bff3e31ab1223fffade0c10971512e660d274d5/galaxy-stable/templates/galaxy_container_cfgmap.yaml#L38

PR #32, opened 2019-02-09
[WIP] Improve 18.01 example to allow tools installation

pcm32:
With this changes the flavour less image should resolve conda packages for tools installed from the toolshed, making the flavour-less setup more interesting for people trying it.


pcm32:
Thanks for reviewing. Unfortunately this had an issue, but since I had it on WIP, I didn't rush to fix it, sorry!... I will revert the merge, fix and merge.

PR #34, opened 2019-02-17
NONUSE slurm

pcm32:
Add ability to set NONUSE on slurm, so that newer 18.09 versions don't need to wait for lengthy slurm startup (2 minutes I think currently). This is not finished or tested yet.


pcm32:
I have tested this and it works well.

nuwang:
LGTM!

PR #37, opened 2019-03-07
Fix rbac to allow multiple deployments

pcm32:
Currently RBAC API objects where using the same name regardless of the helm deployment name, this PR fixes that so that. I have used this locally to work fine.


pcm32:
Thanks @nuwang

PR #43, opened 2019-06-14
Galaxy chart v3

afgane:
We'd like to introduce the so-called v3 of the chart with the following advantages/features:
1. Adhere to Helm best practices;
2. Use of a drastically simplified and smaller Galaxy Docker image (~200MB) with quicker startup time;
3. Decomposed job and web handlers, allowing easier scalability and best practice in terms of how a modern Galaxy should be configured;
4. Implements CVMFS as a per node daemonset, with per node caching for performance via CSI;
5. Unifies handling of Galaxy configs. Instead of exposing a subset of Galaxy configs as individual environment variables, it allows all Galaxy configs to be handled in a uniform way;
6. Minimizes customised dependencies and uses stock charts (e.g., Postgres);
7. Minimizes the need to build custom images, because the integration with the CVMFS for incorporation of the tools, and all of the configs can be changed without a rebuild;
8. Gives a basic deployment model and a CVMFS-based one, which largely replicates the functionality of [Galaxy Main](https://usegalaxy.org/).
9. Tightens security of jobs

The basic set of chart values can be be tested locally and the CVMFS integration can be tested via [CloudLaunch on AWS](https://launch.usegalaxy.org/catalog/appliance/cloudman-20).

Making this PR against the `master` branch, for the reasons described in https://github.com/galaxyproject/galaxy-kubernetes/issues/42.


pcm32:
Is this WIP or ready to merge? For many of our clusters, we pre-provision a PVC which we then use with the branch, however I think I saw the functionality to use a pvc-name gone (I might be wrong though). If that is the case, could you add this back to ease my testing? Thanks!

nuwang:
The current development branch has been tagged as v2.x for posterity, so it'll continue to be accessible. I'll merge this now and set v3 master as the default branch, as we will be using this for the training session.

PR #45, opened 2019-07-06
Galaxy job readiness probe

luke-c-sargent:
changed some behavior; python script for server is now executed through bash to pipe to tee which writes the output to a file, `/tmp/log`.

readiness probe checks this log for "is running", a string output when the job service is ready to accept jobs.

let me know what i can do to improve the quality of this PR

### DIFF
--- a/galaxy/templates/deployment-job.yaml
+++ b/galaxy/templates/deployment-job.yaml
@@ -58,8 +58,17 @@ spec:
 {{- end }}
             - name: GALAXY_CONFIG_DATABASE_CONNECTION
               value: postgresql://{{ .Values.postgresql.galaxyDatabaseUser }}:$(GALAXY_DB_USER_PASSWORD)@{{ template "galaxy-postgresql.fullname" . }}/galaxy
-          command: ["python", "/galaxy/server/scripts/galaxy-main"]
-          args: ["-c", "/galaxy/server/config/galaxy.yml", "--server-name", "$(POD_NAME)", "--attach-to-pool", "job-handlers"]
+#          command: ["python", "/galaxy/server/scripts/galaxy-main"]
+          command: ["/bin/bash", "-c", "python /galaxy/server/scripts/galaxy-main -c /galaxy/server/config/galaxy.yml --server-name $(POD_NAME) --attach-to-pool job-handlers 2>&1 | tee /tmp/log"]
+#          args: ["-c", "/galaxy/server/config/galaxy.yml", "--server-name", "$(POD_NAME)", "--attach-to-pool", "job-handlers"]
+          readinessProbe:
+            exec:
+              command:
+                - "/bin/bash"
+                - "-c"
+                - "tail -n 10 /tmp/log | grep \"is running\""
+#            initialDelaySeconds: 30
+#            periodSeconds: 15
           volumeMounts:
             {{- range $key,$entry := .Values.configs }}
             - name: galaxy-conf-files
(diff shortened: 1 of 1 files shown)


galaxy/templates/deployment-job.yaml
```diff
@@ -58,8 +58,13 @@ spec:
...
               value: postgresql://{{ .Values.postgresql.galaxyDatabaseUser }}:$(GALAXY_DB_USER_PASSWORD)@{{ template "galaxy-postgresql.fullname" . }}/galaxy
-          command: ["python", "/galaxy/server/scripts/galaxy-main"]
-          args: ["-c", "/galaxy/server/config/galaxy.yml", "--server-name", "$(POD_NAME)", "--attach-to-pool", "job-handlers"]
+          command: ["/bin/bash", "-c", "python /galaxy/server/scripts/galaxy-main -c /galaxy/server/config/galaxy.yml --server-name $(POD_NAME) --attach-to-pool job-handlers 2>&1 | tee /tmp/log"]
+          readinessProbe:
+            exec:
+              command:
+                - "/bin/bash"
+                - "-c"
+                - "tail -n 10 /tmp/log | grep \"is running\""
```
almahmoud: Tailing only 10 lines could miss the "is running" line if jobs are picked up between when it starts running and the end of a period to re-check:
eg: if a user has 11 jobs queued in the database, the check fails 1 sec before the running line, and by the time it checks again, the handler picked up 11 jobs (it might actually be just 6 jobs cause I think it logs a line when the job is picked up and another when it's dispatched to a runner, but regardless X # of jobs that would log 11+ lines). Very edge case and will most likely not happen, but it seems like an unnecessary risk.
Perhaps grep the file directly? Given that it's just a readiness probe and it's running only once at the beginning, the file should never be incredibly big, so it should be fine.

almahmoud: Tailing only 10 lines could miss the "is running" line if jobs are picked up between when it starts running and the end of a period to re-check:
eg: if a user has 11 jobs queued in the database, the check fails 1 sec before the running line, and by the time it checks again, the handler picked up 11 jobs (it might actually be just 6 jobs cause I think it logs a line when the job is picked up and another when it's dispatched to a runner, but regardless X # of jobs that would log 11+ lines). Very edge case and will most likely not happen, but it seems like an unnecessary risk.
Perhaps grep the file directly? Given that it's just a readiness probe and it's running only once at the beginning, the file should never be incredibly big, so it should be fine.

galaxy/templates/deployment-job.yaml
```diff
@@ -58,17 +58,13 @@ spec:
...
           command: ["/bin/bash", "-c", "python /galaxy/server/scripts/galaxy-main -c /galaxy/server/config/galaxy.yml --server-name $(POD_NAME) --attach-to-pool job-handlers 2>&1 | tee /tmp/log"]
-#          args: ["-c", "/galaxy/server/config/galaxy.yml", "--server-name", "$(POD_NAME)", "--attach-to-pool", "job-handlers"]
           readinessProbe:
             exec:
               command:
                 - "/bin/bash"
                 - "-c"
                 - "tail -n 10 /tmp/log | grep \"is running\""
-#            initialDelaySeconds: 30
-#            periodSeconds: 15
```
almahmoud: I'd suggest leaving the values explicitly declared, even if they are the same as the defaults. Not necessary, but it makes it easier to understand what is happening/when and later change it if necessary.

almahmoud:
Closing in favor of https://github.com/galaxyproject/galaxy-helm/pull/127

PR #49, opened 2019-08-01
Changing required versions

almahmoud:
CSI 1.0 drivers were not released until K8S 1.13, so using version <1.13 would require using the old CVMFS-CSI 0.3.0.
Also I don't see why Helm would have to be 2.13 or above as it works with 2.10,  as we are not using any of the new features of helm to my knowledge so I updated that as well
(I think the numbers were probably inverted accidentally to say K8S 1.10+ and Helm 2.13+ when in fact we've tested on K8S 1.13+ and Helm 2.10+)

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,8 @@ configuration changes, updates, upgrades, and rollbacks.
 
 ## Prerequisites
 
-- Kubernetes 1.10+
-- Helm 2.13+
+- Kubernetes 1.13+
+- Helm 2.10+
 
 You will need a Kubernetes and Helm installation; the easiest option for
 testing and development purposes is to install


README.md
```diff
@@ -22,8 +22,8 @@ configuration changes, updates, upgrades, and rollbacks.
 
 ## Prerequisites
 
-- Kubernetes 1.10+
-- Helm 2.13+
+- Kubernetes 1.13+
+- Helm 2.10+
```
afgane: Can you check if Helm 2.10 has the `--set-file` option? It looks like it was added with 2.10 (https://github.com/helm/helm/pull/3758) but just to make sure. We use that option in the tutorial.

afgane: Yes it does.

PR #50, opened 2019-08-08
Add a link to the documentation page on Galaxy hub.

VJalili:


afgane:
I don't thing we need to remove the TL;DR but just adding a link to the Hub docs.

PR #54, opened 2019-09-02
Added support for specifying resource limits

nuwang:


nuwang:
@almahmoud @afgane We'll need to get this merged asap, since resource limits don't currently apply. One note is that perhaps the container mapper rule + the test could go into GalaxyCloudRunner, so that it's more self contained - what do you think? It may be too later for that now though.

pcm32:
On our setup we also allow the user set resources via sliders in the UI. In our rules, similar to the ones proposed here, we have some merging logic for the destination and user defined resources. It is a pity though that the tool cannot inform to the UI sliders what is the current setting according to its default destination.

nuwang:
@pcm32 This was inspired by your very setup. It's been made slightly more generic so that the "resource sets" can be configurable.

PR #55, opened 2019-09-03
Change galaxy-helm chart to work with Galaxy image built with libyaml uwsgi

nuwang:
xref: https://github.com/galaxyproject/galaxy/issues/8558
xref: https://github.com/CloudVE/galaxy-helm/commit/31434159c830f376b9b2cefeb779271b49121181

### DIFF
--- a/galaxy/templates/configs-galaxy.yaml
+++ b/galaxy/templates/configs-galaxy.yaml
@@ -16,7 +16,10 @@ data:
 {{- end }}
   {{- range $key, $entry := .Values.configs -}}
   {{- if $entry -}}
-  {{- $key | nindent 2 }}: |
-    {{- tpl (tpl $entry $) $ | nindent 4 -}}
-  {{- end }}
-  {{- end }}
+  {{- $key | nindent 4 }}: |
+  {{- $original := (toYaml $entry) -}}
+  {{- $nomultiline := (regexReplaceAll "^^\\s*\\|\\n*" $original "") -}}
+  {{- $nowhitespace := (regexReplaceAll "{{\\s*([^}\\s]+)\\s*}}" $nomultiline "{{$1}}") -}}
+  {{- tpl (tpl $nowhitespace $) $ | trim | nindent 8 -}}
+  {{- end -}}
+  {{- end -}}
--- a/galaxy/values-cvmfs.yaml
+++ b/galaxy/values-cvmfs.yaml
@@ -172,28 +172,30 @@ configs:
             </destination>
         </destinations>
     </job_conf>
-  galaxy.yml: |
+  galaxy.yml:
     uwsgi:
       virtualenv: /galaxy/server/.venv
       processes: 1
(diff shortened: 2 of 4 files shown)


galaxy/values-cvmfs.yaml
```diff
@@ -171,31 +171,28 @@ configs:
...
       processes: 1
       http: 0.0.0.0:8080
-      static-map: /static/style=/galaxy/server/static/style/blue
-      static-map: /static=/galaxy/server/static
-      static-map: /favicon.ico=/galaxy/server/static/favicon.ico
       pythonpath: /galaxy/server/lib
       thunder-lock: true
       manage-script-name: true
-      mount: {{.Values.ingress.path}}=galaxy.webapps.galaxy.buildapp:uwsgi_app()
       buffer-size: 16384
```
nuwang: @almahmoud The code snippet in the configmap for converting to yaml `{{- $original := ($entry | toYaml) -}}` trips up on this line. Therefore, I made this a commandline argument to uwsgi too, but seems wrong.

almahmoud: Fixed and made some more changes in: https://github.com/galaxyproject/galaxy-helm/pull/55/commits/dd3bd3495e9b6b26abac6a2a291180e3e4288e4e

nuwang:
It's great that you got this working in so many scenarios, given all the complications we've run into. A new path may have opened up: https://github.com/galaxyproject/galaxy/issues/8558
We should try switching uwsgi to libyaml in the docker container, and that could end up simplifying things quite a bit. Then we'd have fewer edge cases to handle.

Ideally, we should be able to treat this all in a simple uniform way - whether or not a config file is a single string literal or a nested dict should ideally not matter since it's all valid yaml.

WRT to deleting helm keys, that seems to be supported? https://github.com/helm/helm/pull/2648

nuwang:
@almahmoud Let's work on the libyaml portion today or tomorrow perhaps?

galaxy/templates/configs-galaxy.yaml
```diff
@@ -16,7 +16,36 @@ data:
 {{- end }}
   {{- range $key, $entry := .Values.configs -}}
   {{- if $entry -}}
-  {{- $key | nindent 2 }}: |
-    {{- tpl (tpl $entry $) $ | nindent 4 -}}
-  {{- end }}
-  {{- end }}
+  {{- $key | nindent 4 }}: |
```
nuwang: @almahmoud Now that the libyaml change has been made: https://github.com/CloudVE/galaxy-docker-k8s/pull/11, it seems like we should convert this to the simplest version that works. Wasn't there some version you wrote that handled both but was fairly simple?

almahmoud: Done!

almahmoud: I also rebased everything into a single commit

PR #57, opened 2019-10-07
19.09 working

almahmoud:
Main changes:
-Versioning + image tag to 19.09
-No more copying for editable shed tool conf (it is now set with `shed_tool_config_file` and will be created automatically)

Needs either a new DB dump or the DB migration init container: https://github.com/galaxyproject/galaxy-helm/pull/56

### DIFF
--- a/galaxy/Chart.yaml
+++ b/galaxy/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
-appVersion: "v19.05"
+appVersion: "v19.09"
 description: Chart for Galaxy, an open, web-based platform for accessible, reproducible, and transparent computational biomedical research.
 icon: https://galaxyproject.org/images/galaxy-logos/galaxy_project_logo_square.png
 name: galaxy
-version: 3.0.0
+version: 3.1.0
--- a/galaxy/templates/_helpers.tpl
+++ b/galaxy/templates/_helpers.tpl
@@ -93,5 +93,5 @@ Return galaxy database user password
 Creates the bash command for the init containers used to place files and change permissions in the galaxy pods
 */}}
 {{- define "galaxy.init-container-commands" -}}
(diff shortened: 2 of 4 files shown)


jmchilton:
Screen Shot 2019-10-07 at 1 32 40 PM

Trying to test this and I encountered the chart for the first time and working off this branch - any ideas?

afgane:
Are you running this locally on a Mac? Unless something's changed lately, the CVMFS chart does not work on a Mac and could be a cause. Try setting `cvmfs.enabled = False` in `values.yaml` (https://github.com/galaxyproject/galaxy-helm/blob/d0f9b42b8f6c5e5a6a9c862ffc982a2bbb7c7023/galaxy/values.yaml#L130). Otherwise, the default chart values may have drifted a bit away from local deployment and we'll need to adjust them.

jmchilton:
Locally on a Mac but that is disabled...

```
grep -A 1 cvmfs values.yaml                  1909
cvmfs:
  enabled: false
```

almahmoud:
> Are you running this locally on a Mac? Unless something's changed lately, the CVMFS chart does not work on a Mac and could be a cause. Try setting `cvmfs.enabled = False` in `values.yaml` (
> 
> https://github.com/galaxyproject/galaxy-helm/blob/d0f9b42b8f6c5e5a6a9c862ffc982a2bbb7c7023/galaxy/values.yaml#L130
> 
> ). Otherwise, the default chart values may have drifted a bit away from local deployment and we'll need to adjust them.

I don't think this is related to CVMFS though. The config files `integrated_tool_panel.xml` and `sanitize_whitelist.txt` are not in the galaxy repository anymore. Should we get them from somewhere else (eg: CVMFS) or start without them as we did for the `shed_tool_conf.xml.sample`?

almahmoud:
Should be fixed now

PR #58, opened 2019-10-07
[Workaround] Docker-in-Docker for mulled containers

almahmoud:
This is not necessarily meant to be merged, but a useful workaround until we figure out how to automate the process of finding combinations of tools not in biocontainers.
Having a docker-in-docker container will allow the job containers to build mulled containers on-the-fly and use them when the images are not found in biocontainers.


nuwang:
This is a far more compact solution than I expected. Useful to have this PR around for future reference.

Also linking in this new PR that will allow us to identify the broken containers and fix them at the source: https://github.com/galaxyproject/galaxy/pull/8741

PR #60, opened 2019-10-23
Default schema update to Galaxy 19.09 model

afgane:


galaxy/values-cvmfs.yaml
```diff
@@ -4,7 +4,7 @@
 
 image:
   repository: galaxy/galaxy
-  tag: 19.09m
+  tag: 19.09-k8s
```
nuwang: @afgane I think the tag name would be better off as a semantic version,  and we name the image galaxy-k8s:19.09 or something. This would allow us to do parsing of the version later if necessary. But we can decide on that later.

afgane: galaxy-k8s:19.09 is definitely better than having k8s as part of the tag. I'll update to that. 

As we automate more of the setup, the tag may end up being more like `19.09-date` for automated daily builds for example, for which the tag you suggested will work better. Ultimately, we'll track Galaxy releases so I'm not sure how we'd go about semantic versioning? I guess our main image 'incompatibility' will be when the database models change, but that shouldn't happen within a Galaxy release.

PR #62, opened 2019-10-28
Using zipped configs

almahmoud:
The current approach uses an init container to download and unzip the ~60MB file from https://raw.githubusercontent.com/CloudVE/cvmfs-main-zipped/master/current/zipped.tar.gz (automatically updates daily), which contains the full directories `config/` and `tool_data` from the CVMFS `main.galaxyproject.org` repositories, along with all contents of `shed_tools/` directory minus the directories `.hg`, `test-data`, and `images` omitted anywhere they occur in any tool subdirectory (which amount to about 1.5GB more) and uses postStart lifecycle hooks to add symlinks pointing to the actual CVMFS mounted repository for the files in those omitted directories.
When testing on a new GKE cluster (no CVMFS cache), directly reading configs from the CVMFS has Galaxy starting in ~12 min while using the zipped files cuts it down to ~4 min (roughly estimated from the dashboard)

Some thoughts for future improvement:

- For tools that do not use mulled-containers and rely on a script in the `shed_tools` directory to run, we need to make sure that they have access to those scripts either from the actual CVMFS or the unzipped files. I believe getting them from the actual CVMFS is likely better, and I believe the easiest way to do so would be to not change the pointers inside the tool definition XML files, and leave them pointing to the real CVMFS at `/cvmfs/main.galaxyproject.org`. This would allow the job handler to read from the XML coming from the unzipped directory, but hand off the command pointing to the script on the actual CVMFS (which is already mounted by the job containers). Alternatively, we could use the shared volume to store the unzipped directories (using just an ephemeral volume at the moment), and then we can leave everything pointing to the unzipped directories with symlinks as things are now.
- We might be able to get away with just zipping ~5 MB (by cutting down to only zip `XML` files in the `shed_tools/` directory, and symlink the rest). This should work fine because the rest of the files are generally only needed when running the tool, not for initial loading. This will cut down the time only by ~20 sec when running on GKE, but the difference could be significant if running on a machine with relatively a slow internet connection where downloading 60 MB might take some time.
- I can move the zipped file to automatically push to S3 instead of just GitHub if desired, using CloudBridge to upload the resulting file. Leaving it in GitHub temporarily though as it is easier for development until this is finalized.


galaxy/values-cvmfs.yaml
```diff
@@ -195,12 +195,12 @@ configs:
...
+      tool_config_file: "/galaxy/server/config/tool_conf.xml,/cvmfs/zipped/cvmfs/main.galaxyproject.org/config/shed_tool_conf.xml"
       shed_tool_config_file: "{{.Values.persistence.mountPath}}/config/editable_shed_tool_conf.xml"
-      tool_data_table_config_path: "{{ .Values.cvmfs.main.mountPath }}/config/shed_tool_data_table_conf.xml,{{.Values.cvmfs.data.mountPath}}/managed/location/tool_data_table_conf.xml,{{.Values.cvmfs.data.mountPath}}/byhand/location/tool_data_table_conf.xml"
+      tool_data_table_config_path: "/cvmfs/zipped/cvmfs/main.galaxyproject.org/config/shed_tool_data_table_conf.xml,{{.Values.cvmfs.data.mountPath}}/managed/location/tool_data_table_conf.xml,{{.Values.cvmfs.data.mountPath}}/byhand/location/tool_data_table_conf.xml"
       tool_dependency_dir: "{{.Values.persistence.mountPath}}/deps"
       builds_file_path: "{{.Values.cvmfs.data.mountPath}}/managed/location/builds.txt"
-      datatypes_config_file: "{{ .Values.cvmfs.main.mountPath }}/config/datatypes_conf.xml"
+      datatypes_config_file: "/cvmfs/zipped/cvmfs/main.galaxyproject.org/config/datatypes_conf.xml"
```
nuwang: Are there any tools that use libgalaxy and read the datatypes_config_file? If so, the tool would not have the zipped directory?

galaxy/templates/deployment-web.yaml
```diff
@@ -57,6 +59,10 @@ spec:
         - name: {{ .Chart.Name }}-web
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          lifecycle:
+            postStart:
+              exec:
+                command: ['/bin/sh', '-c', 'find /cvmfs/main.galaxyproject.org/shed_tools -type f -path "*/.hg/*" -exec cp -s --parents \{\} /cvmfs/zipped/ \; && find /cvmfs/main.galaxyproject.org/shed_tools -type f -path "*/test-data/*" -not -path "*/.hg/*" -exec cp -s --parents \{\} /cvmfs/zipped/ \; && find /cvmfs/main.galaxyproject.org/shed_tools -type f -path "*/images/*" -not -path "*/test-data/*" -not -path "*/.hg/*" -exec cp -s --parents \{\} /cvmfs/zipped/ \; ']
```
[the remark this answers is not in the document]
almahmoud: If we stick with ephemeral storage like now, then we'd need it in both and then templating it like the init container command would be nice. If we however switch to a shared volume, it won't be necessary anymore as only one of the deployments would have to run it.

galaxy/values-cvmfs.yaml
almahmoud: If you by "use libgalaxy" you mean not have a mulled container and just use the minimal image fallback, yes, we still need to mount the zipped volume into job pods

almahmoud:
I'm going to close this but leave the branch in case it's needed in the future. With the alien cache enabled and preloading introduced in CVMFS-CSI Chart (v1.3), this is not needed. There is a single cache on the NFS shared by the CVMFS drivers running on each node, and there are non-blocking kubernetes jobs that are deployed by the helm chart on install to pre-cache certain files (eg: `cat` xml wrappers). Galaxy startup is ~3-3.5 minutes from the first run (about same as restarts)

PR #83, opened 2019-12-15
Create a dedicated workflow handler deployment

afgane:
Related to https://github.com/galaxyproject/galaxy/issues/9040. The number of replicas of the workflow handler needs to be kept at 1 for the time being because the `workflow_schedulers_conf.xml` requires an apriori named handler.

### DIFF
--- a/galaxy/templates/deployment-job.yaml
+++ b/galaxy/templates/deployment-job.yaml
@@ -69,7 +69,7 @@ spec:
         {{- range $each := .Values.extraInitContainers -}}
         {{- if $each.applyToJob -}}
         {{- print "- " | nindent 8 -}}
-        {{- tpl ((unset (unset (unset $each "applyToJob") "applyToWeb") "applyToWorkflow") | toYaml | indent 10 | trim) $ -}}
+        {{- tpl ((unset (unset $each "applyToJob") "applyToWeb") | toYaml | indent 10 | trim) $ -}}
         {{- end }}
         {{- end }}
         {{- end }}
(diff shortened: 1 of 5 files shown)


nuwang:
> I was guessing you were going to say something like that. However, while the structure is largely repeated 3 times, it is only 3 times. I don't think we'll ever have 4 so it's capped. Meanwhile, it is more explicit to have the full content in the file one is looking at and hence more accessible to potential (new) contributors instead of turning into a framework. I'd suggest we leave it like this for the time being and reevaluate in the future.

Sure, you concern is valid. I wasn't thinking of moving everything en masse to _helpers.tpl. Only repetitive elements which do not hinder comprehension of the structure. This is for two reasons:
 - It's already the case that a person cannot contribute to this chart without having to use templates, so a template include here and there is not difficult to comprehend.
- We are already failing to keep 2 copies in sync (e.g. we can't keep values.yml and values-cvmfs in sync), so keeping 3 things in sync is going to be that much harder. 

So if we do this this cleverly, we can reduce repetition without affecting comprehensibility. On the other hand, a wrong thing here or there could lead to subtle, difficult to diagnose errors. If you feel strongly about it in one way or the other though, I don't mind, we can give this a go and see.

nuwang:
> I added the schedulers file instead realizing that the workflow schedulers deployment will run regardless of which values file is used.

It seems like we are implicitly overriding the default schedulers conf file? Everywhere else, we've gone for an explicit approach.

nuwang:
> > I added the schedulers file instead realizing that the workflow schedulers deployment will run regardless of which values file is used.
> 
> It seems like we are implicitly overriding the default schedulers conf file? Everywhere else, we've gone for an explicit approach.

Just realized that the original line is intact, so this is ok.

afgane:
Removed the `applyToWorkflow` addition for init scripts and containers in favor of reusing the ones defined for the job. No scenario comes to mind where the two processes should be different so simplifying it makes sense.

PR #84, opened 2019-12-15
Add extraFileMappings and extraInitCommands

nuwang:
This PR does the following:
1. Adds extraFileMappings to allow mapping arbitrary files into the docker container
2. Adds a default welcome.html to the container using extraFileMappings
3. Deprecates extraInitScripts in favour of extraInitCommands, which is a single script. That script can call additional init scripts mounted through extraFileMappings. It can also conditionally do so if desired by checking whether the relevant file is mapped.


nuwang:
I was thinking the other way around is better? That's because the workflow is a copy of the job deployment right? So could we just copy the job deployment again and change the command?

afgane:
Since we merged #83 first, this PR will now need to be updated to include the changes to the workflow deployment as well.

PR #87, opened 2019-12-20
Add static-safe mapping to client/galaxy/images

dannon:
This fixes various fugue/etc icons not displaying correctly (https://github.com/galaxyproject/galaxy/issues/9080\#issuecomment-564590700)


nuwang:
Thanks @dannon

PR #89, opened 2020-01-07
DB Password

almahmoud:
Main changes:
- If db password is not specified, it will be auto-generated
- Regardless of whether auto-generated or specified, it will be stored in a secret
- Adding capabilities to specify an existing secret & key for the password via `galaxyExistingSecret` and `galaxyExistingSecretKeyRef`


nuwang:
Commits included with: https://github.com/galaxyproject/galaxy-helm/pull/88

PR #90, opened 2020-01-08
Update model and defaults to 20.01

afgane:
This updates the default model to Galaxy 20.01 along with the default image values. The version of Postgres chart is also updated to the latest version (the model was built using the new version).


mvdbeek:
Just a heads up, you don't need to include a postgres dump (assuming that's for faster startups) since https://github.com/galaxyproject/galaxy/pull/8300 / 19.09

afgane:
Just confirmed this working as advertised @mvdbeek! Thanks for letting us know.

nuwang:
@afgane The web pods seem to be ok, but wanted to check on the job pods. I believe they are applying the migrations one at a time.

afgane:
@mvdbeek on the second pass, as Nuwan noticed, a different pod was in fact running through all the migrations. It seems the Galaxy code assumes it will be creating a new database altogether: https://github.com/mvdbeek/galaxy/blob/3f4bef05f7a5dd834a0e326284b6a27afaf0a7e0/lib/galaxy/model/migrate/check.py#L35 (and https://sqlalchemy-utils.readthedocs.io/en/latest/database_helpers.html#database-exists). This doesn't work in our case because we pre-create a database (https://github.com/galaxyproject/galaxy-helm/blob/6b145d37444ca6a8e57a52cfe7fcbec32f7ca562/galaxy/files/conf.d/init1_dbobjects.sh#L6) and set suitable permissions (can't really allow the `galaxy` user complete control over the database).

Would it be possible to modify the Galaxy behavior to also check if the database is empty and then skip applying all the migrations?

mvdbeek:
I'll have a look, that should be possible as well.

afgane:
@mvdbeek Please let us know if/when an empty database will work and we'll remove the model dump but given this is likely to take a while, I've just included the dump for the time being.

PR #91, opened 2020-01-09
Add a basic tool_conf to values

afgane:
A tool conf specified in `galaxy.yml` was missing from the config.


nuwang:
Should we just change it to use the sample instead?

PR #92, opened 2020-01-14
Added pod priority to job spec

nuwang:


nuwang:
The idea was to generally ignore the priority. The preemptionPolicy=never is what does the work here.

nuwang:
Related PR: https://github.com/galaxyproject/galaxy/pull/9219

PR #98, opened 2020-01-22
Fix ordering of env vars

nuwang:
xref: https://github.com/kubernetes/kubernetes/issues/40373


afgane:
Turns out this is an issue with Rancher UI so no need for this change: https://github.com/rancher/rancher/issues/16148

PR #99, opened 2020-01-22
Add attribute named "optional" so rancher doesn't try to update the entry on redeployment

nuwang:


afgane:
Turns out this is an issue with Rancher UI so no need for this change: https://github.com/rancher/rancher/issues/16148

PR #104, opened 2020-01-31
Add larger resource sets and associate tools

afgane:
Used this list as an inspiration but without actually taking the CPU values into account: https://github.com/usegalaxy-eu/sorting-hat/blob/master/tool_destinations.yaml

It's a first pass, better than nothing but will need ongoing tweaking. ~~Also, this includes only the latest versions of the select tools, which I think is OK for our application.~~ 

Purposefully didn't update `values.yaml` as that's primarily intended to be run locally so increasing resource reqs doesn't make too much sense to me.

### DIFF
--- a/galaxy/values-cvmfs.yaml
+++ b/galaxy/values-cvmfs.yaml
@@ -482,36 +482,33 @@ jobs:
             docker_container_id_override: {{ .Values.image.repository }}:{{ .Values.image.tag }}
             resource_set: small
         - tool_ids:
-            - toolshed.g2.bx.psu.edu/repos/devteam/bwa/bwa/0.7.17.4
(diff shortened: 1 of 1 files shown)


afgane:
Something else that's come to my attention looking into this is that the job / job runner ignores the `request` for the number of CPUs/threads and instead use the `limit` value. Should we hence just remove the `request` value from the `resource_sets`? 

As it stands, K8s thinks the jobs run with the `request` value but the command is actually constantly running with the `limit` value so we're oversubscribing everything by a factor of 2 by default. This may be ok if a process is not using the CPUs 100% of the time but a factor of 2 may be too much. Perhaps `n-1`?

Issue #106, opened 2020-02-03
Ingress & PriorityClass versions are not compatible with k8s 1.13


pcm32:
Using current master. I'm running with the following versions:

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2", GitCommit:"59603c6e503c87169aea6106f57b9f242f64df89", GitTreeState:"clean", BuildDate:"2020-01-23T14:21:36Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.5", GitCommit:"2166946f41b36dea2c4626f90a77706f426cdea2", GitTreeState:"clean", BuildDate:"2019-03-25T15:19:22Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}

$ helm version
Client: &version.Version{SemVer:"v2.13.1", GitCommit:"618447cbf203d147601b4b9bd7f8c37a5d39fbb4", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.13.1", GitCommit:"618447cbf203d147601b4b9bd7f8c37a5d39fbb4", GitTreeState:"clean"}
```

Using a values file for my settings:

```
webHandlers:
  replicaCount: 3

jobHandlers:
  replicaCount: 1

persistence:
  existingClaim: galaxy-pvc-nfs

service:
  type: NodePort
  port: 30700
```

On the values file I have additional configs for galaxy.yaml, but I doubt that they would have any impact.

When executing inside the galaxy directory:

```
helm install --values galaxy-config-helmv3.yaml .
```

I get:

```
Error: validation failed: [unable to recognize "": no matches for kind "Ingress" in version "networking.k8s.io/v1beta1", unable to recognize "": no matches for kind "PriorityClass" in version "scheduling.k8s.io/v1"]
```

Thanks!

pcm32:
Ok, Ingress moved to `v1beta1 networking.k8s.io` on 1.14, so documentation is wrong and 1.13 is no longer supported.

pcm32:
PriorityClass move to v1 on 1.15. So, question is, do we really need to move forward on these two to leave 1.13 behind? 

PriorityClass v1beta1 would allow backward compatibility to 1.13. 
Ingress v1beta1 extensions would allow backward compatibility to 1.13.

[...]

PR #107, opened 2020-02-03
Backdates API versions of Ingress and PriorityClass for 1.13 compatibility

pcm32:
This PR re-enables backward compatibility with k8s 1.13 (.5 tested), addressing #106


pcm32:
Thanks for testing @almahmoud!

PR #113, opened 2020-02-12
Makes nodePort port configurable

pcm32:
This PR adds support for NodePort port to be configured through `service.ports[*].nodePort`. I have tested this to work in adequately changing the service when NodePort is set. It adds the default that we normally used on previous versions (which will make all my previous docs on galaxy kubernetes spread around the internet less outdated).


pcm32:
Thanks @afgane!!

PR #116, opened 2020-02-18
Do not create dynamic pvc if an existing one is set for Galaxy

pcm32:
This PR fixes a bug where the dynamic pvc for Galaxy is still created when an existing PVC is set to be used.


nuwang:
Nice catch!

PR #122, opened 2020-02-26
Initial testing additions for spin up of chart

pcm32:
This PR adds logic for spinning up the chart as part of the tests. It is based on an older setup, so possibly some additional things could be removed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ script:
      done
   # Show pod logs/describe if there is an error
   - >
-     bash util/wait-for-deployment -n default -t 600 $(kubectl get deployment | awk '{ print $1 }' | grep 'galaxy-web$')
+     bash util/wait-for-deployment -n default -t 600 $(kubectl get deployment | awk '{ print $1 }' | grep 'galaxy-web$');
      if [ $? != 0 ]; then
        for pod in $(kubectl get pods | awk '{ print $1 }' | grep -v postgres | grep galaxy); do
          kubectl logs $pod
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ install:
   - wget -q ${HELM_URL}/${HELM_TGZ}
   - tar xzfv ${HELM_TGZ}
   - PATH=`pwd`/linux-amd64/:$PATH
-  - helm init --client-only
 
 before_script:
   - sudo mount --make-rshared /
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,24 +1,56 @@
...
+  - kubectl cluster-info
+  - helm dependency update ./galaxy
+  - helm install -f helm-configs/.travis-deployment.yaml --name travis ./galaxy
+  # Check charts format in the meantime.
   - >
      for dir in galaxy; do
       helm lint $dir
-      if [ $? != 0 ]; then
-       travis_terminate 1
-      fi
```
pcm32: actually, I think that we should leave this... no point in running things if the linting fails.

.travis.yml
```diff
@@ -1,24 +1,59 @@
...
 install:
   # Installing Helm
   - wget -q ${HELM_URL}/${HELM_TGZ}
   - tar xzfv ${HELM_TGZ}
   - PATH=`pwd`/linux-amd64/:$PATH
-  - helm init --client-only
+
+before_script:
+  - sudo mount --make-rshared /
+  - sudo apt-get install -y socat
```
nuwang: I don't see socat being called anywhere. Is this a minikube dependency?

.travis.yml
```diff
@@ -1,24 +1,59 @@
...
      done
+  # Describe pods as they are sent to run.
+  - >
+     for pod in $(kubectl get pods | awk '{ print $1 }' | grep -v postgres | grep galaxy); do
+      kubectl describe pods/$pod
+     done
+  - bash util/wait-for-deployment -n default -t 600 $(kubectl get deployment | awk '{ print $1 }' | grep 'galaxy-web$')
+  # Show pod logs once they are running
+  - >
+     for pod in $(kubectl get pods | awk '{ print $1 }' | grep -v postgres | grep galaxy); do
```
nuwang: I feel like showing the pod logs is not necessary unless there's an error.

.travis.yml
```diff
@@ -1,24 +1,59 @@
...
+      kubectl logs $pod
+     done
+  # Describe pods as they are running.
+  - >
+     for pod in $(kubectl get pods | awk '{ print $1 }' | grep -v postgres | grep galaxy); do
+      kubectl describe pods/$pod
+     done
+  - kubectl get pods
+  - kubectl get deployments
```
nuwang: Was there a reason to prefer deployment rollout status over polling a Galaxy end point, and maybe running a small test job? The latter would encompass the former?

pcm32:
Once you approve the PR I can squash and merge.

pcm32:
I think that in general we should do that for all our PR merges from now on, gives a cleaner history. I do that in most of our relevant repos.

.travis.yml
pcm32: Simply that I copied and adapted an older setup that I had at another repo. There is certainly room for improvement, like what you propose which would be great to have, hopefully in subsequent PRs :-).

.travis.yml
pcm32: well, I haven't tried it now, but it used to be due to an issue between helm and minikube. Not sure on current versions.

pcm32:
I think I have addressed most of your very good suggestions @nuwang, if you're happy with this I can squash-merge. I think it would be great to add a tool execution later.

almahmoud:
Actually, would it make sense to add a CVMFS-enabled test here as well?

pcm32:
Oops... missed your comment @almahmoud.... sure, we could do that eventually. We could maybe have some parallel branching tests for different use cases.

nuwang:
> I think that in general we should do that for all our PR merges from now on, gives a cleaner history. I do that in most of our relevant repos.

The best option for a clean history is an interactive rebase. Squashed merges only make sense when the PR has related changes. It wouldn't make sense to squash merge an entire version change for example, since it contains multiple unrelated commits, which makes a single commit impossible to comprehend in isolation. Deciding on a case by case basis is best I think.

pcm32:
> Squashed merges only make sense when the PR has related changes.
This is what should happen most of the times. Otherwise, you're being naughty in the PR. Release/version changes branches are completely different things, but for most feature/bug fixes PRs, I would expect squashed merges.

PR #124, opened 2020-02-29
Reorganized test folders, activated helm tests and removed unused

nuwang:
Documents some commands (e.g. why socat is needed)
Used `helm --wait` to simplify
Run helm tests since we already have a basic test.


.travis.yml
```diff
@@ -26,33 +28,20 @@ before_script:
...
-     for pod in $(kubectl get pods | awk '{ print $1 }' | grep -v postgres | grep galaxy); do
-      kubectl describe pods/$pod
-     done
-  # Show pod logs/describe if there is an error
-  - >
-     bash util/wait-for-deployment -n default -t 600 $(kubectl get deployment | awk '{ print $1 }' | grep 'galaxy-web$');
-     if [ $? != 0 ]; then
+  - helm install --wait --timeout 300 -f ci/values-travis.yaml --name travis ./galaxy
+  - |
+     helm test travis;
```
pcm32: Interesting, wasn't aware of helm test. Does it have the same effect of checking that the deployments are up?

nuwang: It's better in some ways because it actually contacts the service: https://github.com/galaxyproject/galaxy-helm/blob/09cd5d1a67ff5522930e6326d2330f905bf02689/galaxy/templates/tests/test-connection.yaml#L17

We should enhance that test to run a simple test job, or better yet, a simple workflow, at which point it'll be a pretty comprehensive test of all 3 deployments. I'll work on that in the coming week.

nuwang:
@pcm32 Thanks for the review. Regarding additional configs, yes, separate from the test related stuff sounds good.

PR #126, opened 2020-03-12
Make dbinit optional and migration tips.

pcm32:
This PR:

- Makes the dbinit part optional (for cases where you are re-using existing PVCs for postgresql that already have the database).
- Adds some pointers on the doc for migration cases.

### DIFF
--- a/galaxy/templates/secret-initdb.yaml
+++ b/galaxy/templates/secret-initdb.yaml
@@ -1,3 +1,4 @@
+{{ if .Values.postgresql.initdbScriptsSecret }}
 apiVersion: v1
 kind: Secret
 type: Opaque
@@ -13,3 +14,4 @@ stringData:
   {{ range $path, $bytes := .Files.Glob "files/conf.d/*" }}
     {{- base $path }}: {{ tpl ($root.Files.Get $path) $ | quote }}
   {{ end }}
+{{ end }}
--- a/README.md
+++ b/README.md
@@ -169,7 +169,21 @@ volume.
 
 ## Note about persistent deployments and restarts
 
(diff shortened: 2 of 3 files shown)


galaxy/templates/secret-initdb.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if .Values.postgresql.initdbScriptsSecret }}
```
nuwang: Instead of disabling init, I think it would be better if we made the init script rerunnable.  We'd just need to add an `IF NOT EXISTS` to this line and the next: https://github.com/galaxyproject/galaxy-helm/blob/5e2508beb87b973ebaa282f5fb9fbd60ab137b9b/galaxy/templates/secret-initdb.yaml#L12-L13 
This would also synchronise the randomly generated password for the galaxy database, although we'd still need to do the steps you detailed above for the master RDBMS password.

pcm32: Well, I'm not sure if you want to do any of those steps if the database is already in place, but I agree that a CREATE IF NOT EXISTS would be good in addition. I would keep the conditionality as well because, if for whatever reason, the user wouldn't want to use this secret, setting `postgresql.initdbScriptsSecret: null`, which is perfectly possible because this is a dependency chart variable that can be set, the secret would still be created. So I think that this leaves us at a more consistent state.

pcm32: Besides, I already have setup running like this on GKE, so I would appreciate if I can keep the behaviour. It would be more work for me to re-try in a different manner there. And we have people that are using it already.

pcm32:
@nuwang apparently postgres doesn't support IF NOT EXISTS in CREATE DATABASE. Will revert the change now.

pcm32:
Ping @nuwang could we merge this as it is given that IF NOT EXISTS wasn't feasible? Thanks!

nuwang:
@pcm32 Thanks for trying IF NOT EXISTS. I didn't see this ping, sorry I didn't take a look earlier - looks good to me.

PR #127, opened 2020-03-12
Job probes

luke-c-sargent:
Hello;

This PR seeks to add liveness / readiness checks to the job handler pods; should this approach be deemed acceptable, I am fairly certain this can be extended to the workflow scheduler pod as well (and web handlers for that matter, though the http approach is probably a more relevant test).

The approach is based on the observation that the `worker_process` database table contains heartbeat data for all galaxy-associated processes (e.g., web, job and workflow handlers); this heartbeat happens every 60 seconds and simply updates a timestamp. by checking the current time against this time stamp and ensuring that the difference is <= 60s, we can determine that the process is 'live.' fortunately, galaxy dependencies include a python library for communicating with postgres DBs (psycopg2) which is being used here to avoid extra dependencies.

this is added to values files:
```
  # Probe variables
  readinessProbe:
    interval: 60
    initialDelaySeconds: 60
    periodSeconds: 60
    failureThreshold: 10
    timeoutSeconds: 3
  livenessProbe:
    interval: 60
    initialDelaySeconds: 60
    periodSeconds: 60
    failureThreshold: 10
    timeoutSeconds: 3
```

these initial defaults work for the time being, but could probably be tweaked some. ideally, we'd include startupProbes to tighten things up, but we need to have k8s 1.16+ available for that. 

these values are then used in probes (e.g., liveness shown here):
```
          livenessProbe:
            exec:
              command: [
                'sh', '-c',
                'python /tmp/probedb.py -e $GALAXY_CONFIG_OVERRIDE_DATABASE_CONNECTION -o $POD_NAME -i {{ $.Values.jobHandlers.livenessProbe.interval }}'
              ]
            initialDelaySeconds: {{ $.Values.jobHandlers.livenessProbe.initialDelaySeconds }}
            periodSeconds: {{ $.Values.jobHandlers.livenessProbe.periodSeconds }}
            failureThreshold: {{ $.Values.jobHandlers.livenessProbe.failureThreshold }}
            timeoutSeconds: {{ $.Values.jobHandlers.livenessProbe.failureThreshold }}
```
Notes: 
 - in the case of the job handler pod (and possibly workflow handler pod), communication is done through a shared database. Since the pod will be accepting no traditional traffic, is a readiness probe even appropriate here? being live and not 'ready' (e.g., available to work but not accepting traffic) is not detrimental because it only needs database access, which the liveness probe already checks.
 - if/when kubernetes 1.16+ is available on GKE (etc.) for use with `galaxy-helm`, a startupProbe would really help clean up the other probes who have to be resilient enough to be a meaningful check for a healthy container but also tolerate the long startup time. With a startup probe, we can increase the specificity of the liveness probes.

thanks for reading, please let me know what i can do to improve the quality of the submission. if this is a reasonable approach, my next steps (after implementing suggested changes) would be applying this approach to the other pods, where appropriate.


galaxy/values-cvmfs.yaml
```diff
@@ -95,7 +108,12 @@ extraFileMappings:
               </div>
           </body>
       </html>
-
+  /tmp/probedb.py:
```
almahmoud: This should be added to non-cvmfs values file as well if we keep it in values. Alternatively, we could add it in directly in the template since it's essential to the probes?

galaxy/scripts/probedb.py
```diff
@@ -0,0 +1,83 @@
...
+
+DEBUG=False
+# allowable seconds since db timestamp
+default_interval=60
+
+def dprint(msg):
+    if DEBUG:
+        print(msg)
+
+parser = argparse.ArgumentParser()
```
nuwang: It would be good to define a main function and move the code there, following this example: https://github.com/galaxyproject/galaxy/blob/6da7b1ade8866dd9048881c20e37815a79e50f89/scripts/api/fetch_to_library.py#L33
This is a more comprehensive sample:
https://github.com/galaxyproject/galaxy/blob/dev/scripts/cleanup_datasets/cleanup_datasets.py

galaxy/scripts/probedb.py
```diff
@@ -0,0 +1,83 @@
...
+    if len(rows) != 1:
+        dprint("ERROR: there should be only one row, but {} found".format(len(rows)))
+        exit(1)
+    for row in rows:
+        now = datetime.now()
+        less_than_interval(row[1], now, interval)
+
+
+for arg in args:
+    add_arg(**arg)
```
nuwang: It's usually more common to define parser args as follows:
https://github.com/galaxyproject/galaxy/blob/6da7b1ade8866dd9048881c20e37815a79e50f89/scripts/cleanup_datasets/cleanup_datasets.py#L76
This is because parser args are often not this uniform, and it's common to have additional parameters.

galaxy/scripts/probedb.py
```diff
@@ -0,0 +1,83 @@
...
+if parsed_args.verbose:
+    DEBUG=True
+
+if parsed_args.environmentvariable:
+    connection_string = parsed_args.environmentvariable
+else:
+    connection_string = "dbname='" + parsed_args.dbname + "' user='" + parsed_args.username + "' host='" + parsed_args.dbhost + "' password='" + parsed_args.password + "'"
+query_string = """SELECT server_name, update_time FROM worker_process WHERE hostname='""" + parsed_args.hostname + "';"
+
+dprint("provided arguments: \n" + str(parsed_args))
```
nuwang: It would be better to use log.debug here.

galaxy/scripts/probedb.py
```diff
@@ -0,0 +1,83 @@
...
+]
+
+
+def check_rows(connection_string, query_string, interval=60):
+    def less_than_interval(a, b, interval=60):
+        if (b-a).seconds > interval:
+            dprint("time delta {} > {}s".format((b-a).seconds, interval))
+            sys.exit(1)
+        dprint("time delta {} <= {}s".format((b-a).seconds, interval))
+        sys.exit(0)
```
nuwang: I think handling the exit code globally would be better:
```
if __name__ == '__main__':
    sys.exit(main())
```
That way, we have a main function that always has a return value.

galaxy/scripts/probedb.py
```diff
@@ -0,0 +1,83 @@
...
+
+    conn = psycopg2.connect(connection_string)
```
nuwang: I think this direct connection makes sense, since we want it to run as fast as possible. Just noting it because most galaxy scripts construct an app object and get Galaxy to handle the connection, but that doesn't seem appropriate here.

galaxy/values-cvmfs.yaml
nuwang: +1 to @almahmoud's suggestion. We have a limited budget for the length of values.yaml, so it would be better to include this as a permanent fixture, maybe in this file: https://github.com/galaxyproject/galaxy-helm/blob/5e2508beb87b973ebaa282f5fb9fbd60ab137b9b/galaxy/templates/configmap-extra-files.yaml#L21

luke-c-sargent: suggested revision made in [configmap-extra-files.yaml](https://github.com/luke-c-sargent/galaxy-helm/blob/10e20ed149970e0ab6bdca95192726cca26b4262/galaxy/templates/configmap-extra-files.yaml#L25-L37), with a deployment-job.yaml [volumeMount](https://github.com/luke-c-sargent/galaxy-helm/blob/10e20ed149970e0ab6bdca95192726cca26b4262/galaxy/templates/deployment-job.yaml#L117-L119) and corresponding [volume](https://github.com/luke-c-sargent/galaxy-helm/blob/10e20ed149970e0ab6bdca95192726cca26b4262/galaxy/templates/deployment-job.yaml#L151-L153)

galaxy/scripts/probedb.py
luke-c-sargent: suggested change [implemented](https://github.com/luke-c-sargent/galaxy-helm/blob/10e20ed149970e0ab6bdca95192726cca26b4262/galaxy/scripts/probedb.py#L40)

galaxy/scripts/probedb.py
luke-c-sargent: suggested change [implemented](https://github.com/luke-c-sargent/galaxy-helm/blob/10e20ed149970e0ab6bdca95192726cca26b4262/galaxy/scripts/probedb.py#L44-L62)

galaxy/scripts/probedb.py
luke-c-sargent: suggested change [implemented](https://github.com/luke-c-sargent/galaxy-helm/blob/10e20ed149970e0ab6bdca95192726cca26b4262/galaxy/scripts/probedb.py#L10-L23)

galaxy/scripts/probedb.py
luke-c-sargent: suggested change [implemented](https://github.com/luke-c-sargent/galaxy-helm/blob/10e20ed149970e0ab6bdca95192726cca26b4262/galaxy/scripts/probedb.py#L89-L90)

PR #128, opened 2020-03-13
Fixing job handlers number of deployments

almahmoud:
Without the `---` only the last API definition is deployed, resulting in ALWAYS 1 deployment, just changing the name. This will appropriately scale the deployments based on `jobHandlers.replicaCount`

### DIFF
--- a/galaxy/templates/deployment-job.yaml
+++ b/galaxy/templates/deployment-job.yaml
@@ -175,4 +175,5 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
     {{- end }}
+---
 {{- end }}


afgane:
This may be a needed syntax fix, but conceptually we're stuck running only a single job handler because of a bug in Galaxy. More here https://github.com/galaxyproject/galaxy-helm/pull/105

almahmoud:
My understanding was that the bug is an issue with having multiple replicas within a deployment, because of name changes. If we have multiple deployments with fixed names, as is now true, each handler should come back up as the same name and continue with its jobs, so scaling up is fine and restarts are fine, it's just scaling down while jobs are running that is a problem?

dannon:
@almahmoud Haven't followed the whole thread, but yes, if you have multiple named handlers coming back up with the same fixed names, then you're good to go.  The problem is when:
- you have handlers A, B
- jobs get assigned to both
- B dies
- C comes up as a replacement

If the second handler here comes back up as 'B', then the only real concern is how long it's down (those jobs assigned to it will just stall, until it does, of course), and otherwise it should work fine.

almahmoud:
I think the handler does always come back as B, because they're numbered, so we are fine deploying with multiple handlers (eg: just wanting to have 2 handlers at all times), as Pablo was saying in that thread. We also have a known bug in the kubernetes runner right now that is causing handlers to go into an infinite error loop (working on fixing it), but until that is fixed we probably do want 2 handlers by default in case one errors out and can't recover, so that new jobs still get queued until an admin manually recovers the bad handler.

In any case, the functionality for scaling the number of deployments instead of replicas was already added when the other PR got merged, the problem is that right now the manifests are not being separated, and thus the parser is only deploying the last one. This is just fixing that bug, not adding scaling itself. Currently, the templating renders wrong:
`helm template galaxy-helm/galaxy/ -f galaxy-helm/galaxy/values-cvmfs.yaml --generate-name --set jobHandlers.replicaCount=2 --output-dir handlerstest/`

```
---
# Source: galaxy/templates/deployment-job.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: RELEASE-NAME-galaxy-job-0

SKIPPING MIDDLE

      volumes:
        - name: galaxy-conf-files
          configMap:
            name: RELEASE-NAME-galaxy-configs
        - name: galaxy-job-rules
          configMap:
            name: RELEASE-NAME-galaxy-job-rules
        - name: galaxy-data
          persistentVolumeClaim:
            claimName: RELEASE-NAME-galaxy-pvc
        - name: cvmfs-gxy-main
          persistentVolumeClaim:
            claimName: RELEASE-NAME-galaxy-cvmfs-gxy-main-pvc
        - name: cvmfs-gxy-data
          persistentVolumeClaim:
            claimName: RELEASE-NAME-galaxy-cvmfs-gxy-data-pvc
apiVersion: apps/v1
kind: Deployment
metadata:
  name: RELEASE-NAME-galaxy-job-1

SKIPPING END
```

Since there is no `---` between the first and second deployment, only the second deployment is actually taken into consideration, so we get a single handler named `RELEASE-NAME-galaxy-job-1` when having `jobHandlers.replicaCount=2`, only 1 handler named `RELEASE-NAME-galaxy-job-2` when having`jobHandlers.replicaCount=3`, etc... as opposed to getting the number of handlers specified. The naming bug was circumvented by scaling the number of deployments instead of the number of replicas within a deployment in the previous PR, this just fixes the rendering for that.

almahmoud:
> Requirements.lock file is needed.

Oops, added back

PR #131, opened 2020-03-27
Mount tools dir into job container

nuwang:


pcm32:
Why don't we simply copy to the PVC (as you are doing) and change the galaxy config pointing to the location on the PVC where this is copied, instead of mingling with the PVC mountpoints for the jobs?

pcm32:
or jobs are currently not mounting the entire galaxy PVC?

nuwang:
@pcm32 That's a good idea. Let me try that.

nuwang:
@pcm32 You mean like this? I wish we had a way to specify subpaths when mounting containers, in fact, I could have sworn that was added, but looking at the code now, I don't see that option anymore.

pcm32:
Yep, plus reverting 171 and 180 of the configs.

nuwang:
@pcm32 Thanks. I'll test this out and make sure everything's working.

PR #144, opened 2020-06-23
Add extraVolumes to default job_conf

almahmoud:
PVCs (only, i.e. not configmaps/secrets) mounted through `extraVolumes` and `extraVolumeMounts` will also be added to the default `job_conf.xml` to be mounted in the job containers. While in the actual rendered values the spaces are all removed, I also changed the previous entries in `values-cvmfs.yaml` to be one per line for easier reading, and then added the logic making the indentation explicit for easier reading. Essentially if there is an `extraVolume` entry with a valid `name` and a valid `persistentVolumeClaim.claimName`, and a corresponding `extraVolumeMount` with the same name and a valid `mountPath`, these values will be appended to the `job_conf.xml` as `,{{$volume.persistentVolumeClaim.claimName:$volumeMount.mountPath}}`. Easiest test is uncommenting the `extraVolumes/volumeMounts` and doing a `helm template`. You can see the rendered `job_conf` will properly list all the volumes

### DIFF
--- a/galaxy/values-cvmfs.yaml
+++ b/galaxy/values-cvmfs.yaml
@@ -49,12 +49,12 @@ persistence:
   size: 10Gi
   mountPath: /galaxy/server/database
 
-# extraVolumes: |
+# extraVolumes:
 #   - name: shared-data
 #     persistentVolumeClaim:
 #       claimName: shared-data-pvc
 
-# extraVolumeMounts: |
+# extraVolumeMounts:
 #   - name: shared-data
 #     mountPath: /opt/project/shared-data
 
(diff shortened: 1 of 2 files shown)


galaxy/values-cvmfs.yaml
```diff
@@ -199,7 +199,30 @@ configs:
             <plugin id="local" type="runner" load="galaxy.jobs.runners.local:LocalJobRunner" workers="4" />
             <plugin id="k8s" type="runner" load="galaxy.jobs.runners.kubernetes:KubernetesJobRunner">
               <param id="k8s_use_service_account">true</param>
-              <param id="k8s_persistent_volume_claims">{{ template "galaxy.pvcname" . }}:{{.Values.persistence.mountPath}},{{ template "galaxy.fullname" . }}-cvmfs-gxy-data-pvc:{{ .Values.cvmfs.data.mountPath }},{{ template "galaxy.fullname" . }}-cvmfs-gxy-main-pvc:{{ .Values.cvmfs.main.mountPath }}</param>
+              <param id="k8s_persistent_volume_claims">
+              {{- template "galaxy.pvcname" . -}}:{{.Values.persistence.mountPath}},
+              {{- template "galaxy.fullname" . }}-cvmfs-gxy-data-pvc:{{ .Values.cvmfs.data.mountPath }},
+              {{- template "galaxy.fullname" . }}-cvmfs-gxy-main-pvc:{{ .Values.cvmfs.main.mountPath -}}
+              {{- if .Values.extraVolumes }}
```
nuwang, line 206: I think it would be better to include this as a template. Otherwise, there's too much code in job_conf that hinders readability + makes overriding job_conf difficult.

almahmoud: I had similar thoughts, specifically and specifically following the `deployToJob`, `deployToWeb` model, but have it be to choose either in which of the three deployments and if in job pods, and/or just in all the deployments and the job pods as either 4 or 2 boolean options if that makes sense

nuwang: That sounds good, would make it consistent with the existing properties.

pcm32: I override job_conf, does this have any impact on that? Is there something else that needs to be overriden? Can I suggest that we make a release before adding this? Thanks!

almahmoud: @pcm32 The `patch` label will bump the version and repackage with a new version number. We don't overwrite old versions anymore so you should be able to just point to the old version if needed.
However, this does not change anything in terms of what you override, it just makes the default include the extraVolumes as well in the job_conf.

almahmoud: Resolving this for now, will add the functionality to choose where to mount later on. This isn't an immediate problem given that one can override the job_conf anyway.

PR #152, opened 2020-07-24
Rabbitmq support

pcm32:
This PR adds support for rabbitmq.

- [x] Password in galaxy secret
-  ~~Erlang key in galaxy secret~~ Not needed actually.
- [x] Create vhost/queue a priori - Galaxy does not create it on connection

This is needed for syncing tool installations from the toolshed across handlers and for external access for pulsar nodes.


pcm32:
This is currently failing because the rabbitmq helm chart handles the existing secret [through a template](https://github.com/bitnami/charts/blob/cc3c27c500df620b1fbf32550daa2fc5c32682e5/bitnami/rabbitmq/templates/_helpers.tpl#L66), at which point apparently .Release.name is not defined yet. 

```
helm install --set rabbitmq.enabled=true,galaxy.rabbitmqPassword=m4r1n0s .
Error: YAML parse error on galaxy/charts/rabbitmq/templates/statefulset.yaml: error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{"default (printf \"%!s(MISSING)-galaxy-secrets\" .Release.Name) .Values.galaxyExistingSecret":interface {}(nil)}
```

@almahmoud  @nuwang do you know if there is a way to force the evaluation of .Release.name before that or should we just go with password being set through env vars for rabbit (instead of the secret, at least for the rabbit container)?

almahmoud:
I don't think the problem is the release name (and I don't think you can ever run helm without a release name in the context).
Unless i'm missing something, it's just that the name itself is not run through tpl, making the rendered YAML invalid. https://github.com/bitnami/charts/pull/3222 this should fix it.

almahmoud:
On another note, I'd say we should version dependencies with exact versions, and bump them ourselves (whether manually or via CI-initiated PRs) when we test new packaged versions, since even the most well-maintained charts can have slip-ups that break things, with this specific `rabbitmq` chart specifically having had that happen a bit ago with `extraEnvVars` breaking (see https://github.com/bitnami/charts/pull/2867)

pcm32:
Yes, agreed on fixing numbers @almahmoud, I set .x to see what version it would use while testing, and then I didn't set it back. Thanks for the PR on rabbitmq.

almahmoud:
@pcm32 https://github.com/bitnami/charts/pull/3445 was merged btw, sorry for the delay

pcm32:
This is now working well. I'm not entirely happy with the password handling (it uses a dummy default from values.yaml), but I didn't know how to inject the value from one secret (the rabbitmq password set in the galaxy secret) into the second secret (secret-rabbitmq) so that it goes into the rabbitmq definitions file. Maybe I could fix it with some directions from @almahmoud. But this works. I have left Rabbitmq turned off by default, but I would say that our production recommendation should be to turn it on.

pcm32:
On a helm upgrade trial I get:

```
ROLLING BACK
Error: render error in "galaxy/charts/rabbitmq/templates/NOTES.txt": template: galaxy/charts/rabbitmq/templates/NOTES.txt:138:4: executing "galaxy/charts/rabbitmq/templates/NOTES.txt" at <include "common.erro...>: error calling include: template: galaxy/charts/rabbitmq/charts/common/templates/_errors.tpl:18:48: executing "common.errors.upgrade.passwords.empty" at <fail>: error calling fail:
PASSWORDS ERROR: you must provide your current passwords when upgrade the release
    'auth.erlangCookie' must not be empty, please add '--set auth.erlangCookie=$RABBITMQ_ERLANG_COOKIE' to the command. To get the current value:

        export RABBITMQ_ERLANG_COOKIE=$(kubectl get secret --namespace default prodding-albatross-galaxy-secrets -o jsonpath="{.data.rabbitmq-erlang-cookie}" | base64 --decode)

Error: UPGRADE FAILED: render error in "galaxy/charts/rabbitmq/templates/NOTES.txt": template: galaxy/charts/rabbitmq/templates/NOTES.txt:138:4: executing "galaxy/charts/rabbitmq/templates/NOTES.txt" at <include "common.erro...>: error calling include: template: galaxy/charts/rabbitmq/charts/common/templates/_errors.tpl:18:48: executing "common.errors.upgrade.passwords.empty" at <fail>: error calling fail:
PASSWORDS ERROR: you must provide your current passwords when upgrade the release
    'auth.erlangCookie' must not be empty, please add '--set auth.erlangCookie=$RABBITMQ_ERLANG_COOKIE' to the command. To get the current value:

        export RABBITMQ_ERLANG_COOKIE=$(kubectl get secret --namespace default prodding-albatross-galaxy-secrets -o jsonpath="{.data.rabbitmq-erlang-cookie}" | base64 --decode)
```

Maybe we should add a warning in the readme and also directly set the erlang cookie to some dummy default to avoid breaking upgrades?

pcm32:
Now affected by #160

galaxy/templates/deployment-workflow.yaml
```diff
@@ -39,6 +39,11 @@ spec:
           volumeMounts:
             - name: galaxy-data
               mountPath: {{ .Values.persistence.mountPath }}
+        {{ if .Values.rabbitmq.enabled }}
+        - name: {{ .Chart.Name }}-init-rabbitmq
+          image: alpine:3.7
+          command: ['sh', '-c', 'until nc -z -w3 {{ template "galaxy-rabbitmq.fullname" $ }} 5672; do echo "waiting for galaxy-rabbitmq service"; sleep 1; done;']
```
nuwang, line 45: @pcm32 Does this mean that Galaxy will not startup if rabbitmq is not up?

nuwang (approved):

ci/values-travis.yaml
```diff
@@ -23,6 +23,9 @@ persistence:
 influxdb:
   enabled: false
 
+rabbitmq:
+  enabled: true
```
nuwang, line 27: I think the default should be false?

nuwang (changes requested):
@pcm32 +1 on the idea of setting a default ERLANG pass with documentation that it needs to be set. Alternatively, we could additionally enforce it as follows: https://github.com/CloudVE/cloudman-helm/blob/fa28ee614fddcecc7345e4ca925359b69f92d45c/cloudman/templates/grafana-datasource-secret.yaml#L29

I think rabbitmq should be disabled by default? If it's enabled, then the required check from above would alert the user that the erlang password must be set.

ci/values-travis.yaml
pcm32: This is only for the testing, on the general values, rabbit is disabled. However, when installing tools from the toolshed and having multiple web/job handlers and you don't have a proper message queue (relay only on kombu + database), you see issues in all the handlers syncing new tool installs that require handlers restart. I have seen this across different Galaxy versions (19.05 to 20.01).

galaxy/templates/deployment-workflow.yaml
pcm32: If you set the ampq connection to rabbit and rabbit is not up, then Galaxy fails if rabbit is not there where expected, so we need to wait for it.

ci/values-travis.yaml
pcm32: I think that we should leave it enabled for testing.

nuwang: Sorry, I didn't notice that this was the travis values file. All good!

pcm32: Could you please revert then the request for changes in the review? Thanks. So that we can merge once it has been tested with the new 20.05 container image.

nuwang: I left it as is because the ERLANG cookie needs a fix though right?

nuwang: @pcm32 Is this good for another round of review, or is there more work to be done?

pcm32: sorry, I haven't had time to come back to this.

afgane:
This was implemented with an Operator in #378.

PR #153, opened 2020-07-24
[WIP] Add customVolume option to persistence (for special built-in volumes eg: CephFS)

almahmoud:
This should make what we did manually with @chrisbarnettster replicable with any CephFS and any other type of custom volumes.
Note: this requires `ceph-common` (and I think also `ceph-fuse`) to be installed on all the nodes. More generally, the `mount -t [storageType]` should work on all the nodes, so for things other than Ceph the corresponding libraries should be installed

This needs a k8s runner change as well

### DIFF
--- a/galaxy/values.yaml
+++ b/galaxy/values.yaml
@@ -49,6 +49,16 @@ persistence:
   accessMode: ReadWriteMany
   size: 10Gi
   mountPath: /galaxy/server/database
+  customVolume: {}
+    # cephfs:
+    #   monitors:
+    #   - 192.102.25.11:6789
+    #   - 192.102.25.19:6789
+    #   - 192.102.25.28:6789
+    #   path: /volumes/my-special-volume/galaxy
+    #   secretRef:
+    #     name: ceph-secret
+    #   user: myuser
 
 # extraVolumes:
 #   - name: shared-data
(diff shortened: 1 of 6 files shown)


galaxy/values.yaml
```diff
@@ -49,6 +49,16 @@ persistence:
   accessMode: ReadWriteMany
   size: 10Gi
   mountPath: /galaxy/server/database
+  customVolume: {}
```
nuwang, line 52: What about about using this mechanism instead?
https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/preexisting-pd

almahmoud: https://github.com/CloudVE/projman-helm/blob/master/projman/values.yaml#L30 yeh I was thinking about just adding the same mechanism I added in projman (https://github.com/CloudVE/projman-helm/blob/master/projman/values.yaml#L30) to the galaxy chart itself. It should work for both ceph and rclone. I haven't had a chance to test but I can try it out next week, should be straightforward.

afgane:
Custom volume types should not be supported by the chart but handled separately. For example, projeman chart can be used instead: https://github.com/CloudVE/projman-helm

PR #154, opened 2020-07-26
Migrations as job

almahmoud:
The job will likely be run earlier than the init container.

### DIFF
--- a/galaxy/templates/job-db-migrations.yaml
+++ b/galaxy/templates/job-db-migrations.yaml
@@ -27,7 +27,7 @@ spec:
       containers:
         - name: {{ .Chart.Name }}-db-migrations
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
-          command: ['sh', '-c', '/galaxy/server/manage_db.sh upgrade']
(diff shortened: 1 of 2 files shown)


galaxy/templates/job-db-migrations.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: batch/v1
+kind: Job
```
nuwang, line 2: Very timely because it's probably time we revisited migrations in general.

I guess this is relying on the idea that Galaxy won't start without the proper database version set. Seems fine, but I'm not quite sure I understand why this job would run sooner than the init container? Could it cause the remaining containers to unnecessarily crash loop?

almahmoud: The job would get dispatched from the beginning, and would be waiting for Postgres to come up along with the Postgres init containers. Another option would be to combine the two init containers in one of the deployments

nuwang: I can't remember what the outcome of the discussion on this was?

almahmoud:
Closing in favor of https://github.com/galaxyproject/galaxy-helm/pull/271

Issue #160, opened 2020-08-30
Start up error for job handler container only


pcm32:
Using the current ci values on minikube I'm getting this error on start only for the job handler deployment:

```
galaxy.jobs DEBUG 2020-08-30 11:24:14,481 Loading job configuration from /galaxy/server/config/job_conf.xml
galaxy.jobs ERROR 2020-08-30 11:24:14,481 Problem parsing the XML in file /galaxy/server/config/job_conf.xml, please correct the indicated portion of the file and restart Galaxy. '>=' not supported between instances of 'NoneType' and 'tuple'
Traceback (most recent call last):
  File "/galaxy/server/lib/galaxy/jobs/__init__.py", line 333, in __init__
    self._configure_from_dict(job_config_dict)
  File "/galaxy/server/lib/galaxy/jobs/__init__.py", line 372, in _configure_from_dict
    self._set_default_handler_assignment_methods()
  File "/galaxy/server/lib/galaxy/web_stack/handlers.py", line 148, in _set_default_handler_assignment_methods
    self.app.application_stack.init_job_handling(self)
  File "/galaxy/server/lib/galaxy/web_stack/__init__.py", line 117, in init_job_handling
    self._init_job_handler_assignment_methods(job_config, base_pool)
  File "/galaxy/server/lib/galaxy/web_stack/__init__.py", line 84, in _init_job_handler_assignment_methods
    self._set_default_job_handler_assignment_methods(job_config, base_pool)
  File "/galaxy/server/lib/galaxy/web_stack/__init__.py", line 537, in _set_default_job_handler_assignment_methods
    if ((dialect.name == 'postgresql' and dialect.server_version_info >= (9, 5))
TypeError: '>=' not supported between instances of 'NoneType' and 'tuple'
Failed to initialize Galaxy application
```

I think that I started seeing this error on rebase to the latest master.

nuwang:
I've also run into this issue and I can't quite understand the root cause either. The quick fix is to have a null check:
```
if ((dialect.name == 'postgresql' and (dialect.server_version_info or (9, 5)) >= (9, 5))
```
However, that doesn't answer the question of why it's null in the first place - is it some sqlalchemy change? @natefoo or @mvdbeek?

mvdbeek:
Don't know what changed, but `server_version_info` seems to only be available if a connection has been made previously.
```
diff --git a/lib/galaxy/web_stack/__init__.py b/lib/galaxy/web_stack/__init__.py
[...]

PR #162, opened 2020-09-23
Web probes get /api/version

almahmoud:
Getting root path for Galaxy would sometimes cause errors (see https://gist.githubusercontent.com/almahmoud/c1e32962b4a931d96c94ffb941ea15eb/raw/aa85624eb4abc3bbcb506dbe7b606fe70c666710/log-web-alexo-anvil-sep-22.log). Since the probes determine whether the handler can receive requests, the intermittent error would cause the handler to be temporarily marked unhealthy, making the GKE Load Balancer stop requests from hitting the handler, making Galaxy intermittently seem down for a little bit of time.
@dannon suggested it might be due to the size of what needs to be rendered for the Galaxy homepage, and to check the `/api/version` endpoint instead. This seems to have fixed the intermittent issues on GKE.

### DIFF
--- a/galaxy/templates/deployment-web.yaml
+++ b/galaxy/templates/deployment-web.yaml
@@ -83,13 +83,13 @@ spec:
           args: []
           readinessProbe:
             httpGet:
-              path: {{ template "galaxy.add_trailing_slash" .Values.ingress.path }}
+              path: {{ template "galaxy.add_trailing_slash" .Values.ingress.path }}/api/version
               port: 8080
             initialDelaySeconds: 30
             periodSeconds: 15
           livenessProbe:
             httpGet:
-              path: {{ template "galaxy.add_trailing_slash" .Values.ingress.path }}
+              path: {{ template "galaxy.add_trailing_slash" .Values.ingress.path }}/api/version
               port: 8080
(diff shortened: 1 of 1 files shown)


dannon (approved):

nuwang:
@almahmoud CI is failing, could you check?

pcm32:
Also, I think that this branch is still testing against the 20.01 container image. When you move it to 20.05 I would expect that you get into the error I'm seeing in #152 (made an issue #160). I did have a successful run with 20.05, but I suspect that there was a change in the container or something. Also, if you could allow me to restart travis jobs, it would be great, thanks! And we should probably move this to travis-ci.com from current travis-ci.org, but I don't have the privs to do it.

nuwang:
Just adding a note that the reason the liveness probe is failing for the root url appears to be a login redirect returned by Galaxy to Keycloak. Why it's not consistently failing I don't know.

almahmoud:
This came up on GKE without keycloak on a single_user instance for me, and it would return 200 most of the time, but sometimes would have “broken pipe” error

nuwang:
included in: https://github.com/galaxyproject/galaxy-helm/pull/163

PR #163, opened 2020-09-24
Adding probes to workflow handlers + adding startup probes + parametrizing web probes

almahmoud:
-Workflow handlers using the same script designed by Luke for job handlers, probing the heartbeat in the DB.
-Startup probes added so we can choose values for the other probes without worrying about startup.
-Parametrized web handler probes to more easily change values per deployment (eg: if running on a slower cloud)
-Rolled in https://github.com/galaxyproject/galaxy-helm/pull/162 into this as well

### DIFF
--- a/galaxy/templates/deployment-job.yaml
+++ b/galaxy/templates/deployment-job.yaml
@@ -99,16 +99,6 @@ spec:
              {{- end -}}
              python /galaxy/server/scripts/galaxy-main -c /galaxy/server/config/galaxy.yml --server-name job_handler_{{ $handler_num }} --attach-to-pool job-handlers']
           args: []
-          startupProbe:
-            exec:
-              command: [
-                'sh', '-c',
-                'python /tmp/probedb.py -e $GALAXY_CONFIG_OVERRIDE_DATABASE_CONNECTION -o $POD_NAME'
-              ]
-            initialDelaySeconds: {{ $.Values.jobHandlers.startupProbe.initialDelaySeconds }}
-            periodSeconds: {{ $.Values.jobHandlers.startupProbe.periodSeconds }}
-            failureThreshold: {{ $.Values.jobHandlers.startupProbe.failureThreshold }}
-            timeoutSeconds: {{ $.Values.jobHandlers.startupProbe.failureThreshold }}
           readinessProbe:
             exec:
               command: [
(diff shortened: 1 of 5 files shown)


almahmoud:
My proposed defaults:

```
  startupProbe:
    initialDelaySeconds: 30
    periodSeconds: 5
    failureThreshold: 720
    timeoutSeconds: 5
  readinessProbe:
    periodSeconds: 10
    failureThreshold: 12
    timeoutSeconds: 5
  livenessProbe:
    periodSeconds: 10
    failureThreshold: 30
    timeoutSeconds: 5
```
for all containers.

I removed the initial delay for the readiness and liveness probes, since those now only run after startup is guaranteed.

What this means effectively:

For all:
`timeoutSeconds` is the amount of time the probe will wait (i.e. the web handler has 5 seconds to reply to the Get request before it times out and the probe attempt is marked unsuccessful)

Startup:
The max allowed startup time is ~1 hour. We can lower this, but figured it's way more than enough leeway for slow systems. The max possible delay from when the container is ready to when it's marked to have finished startup is 5 seconds.

Readiness/Liveness:
After startup, the readiness probe will trigger right away and the container will be marked ready. Afterwards, the status of the container will be checked every 10 seconds, if the container is not responsive for 12 consecutive probes (~2 min) then the container will be removed from the ready pool (new requests will not be directed to it so, for eg: if running in HA mode with at least 2 replicas, all requests will go to the other handler), but the container will not be killed yet. The liveness probe will run another 18 times (for a total of 30 attempts i.e. ~5 min) and if the container is still not responding, it will be marked unresponsive and will be forced to restart. So effectively, if the container is unresponsive because of too much load, it theoretically has ~3 minutes to recover during which new requests are stopped from going to it. If it recovers in this time, it will be marked as ready and continue as normal, if it doesn't it will be killed after 5 minutes of being unresponsive.

TLDR:
Max startup time: ~1 hour
Max unresponsiveness before container is taken out of circulation for new traffic: ~2 min
Max unresponsiveness before container is restarted: ~5 min

Note: These values are largely arbitrary, so very open to suggestions!

dannon (commented):
+1, this looked like it should work great to me.

almahmoud:
Note: startupProbes were introduced as alpha feature in k8s 1.16, and will not work with k8s <= 1.15. I can separate the PRs so we can deal with that?

PR #175, opened 2020-10-26
Build sites with CVMFS removed from non-CVMFS values

almahmoud:


pcm32 (approved):
LGTM, thanks @almahmoud !

PR #184, opened 2020-11-10
New build that includes fix for #160

nuwang:
https://github.com/galaxyproject/galaxy-helm/issues/160#issuecomment-724285438

Thanks @pcm32.

### DIFF
--- a/galaxy/values.yaml
+++ b/galaxy/values.yaml
@@ -4,7 +4,7 @@
 
 image:
   repository: galaxy/galaxy-k8s
-  tag: 20.05
+  tag: 20.05.1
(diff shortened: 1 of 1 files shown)


galaxy/values.yaml
```diff
@@ -4,7 +4,7 @@
 
 image:
   repository: galaxy/galaxy-k8s
-  tag: 20.05
+  tag: 20.05.1
```
pcm32, line 7: Is there are point to keep the non-functional 20.05 (previous image)?

nuwang: In this case, it does make sense yes. Have updated the image.

Issue #174, opened 2020-10-26
CVMFS disabled still leaves breaking config in galaxy.yml


pcm32:
Starting the helm chart with CVMFS disabled still leaves (with chart version 3.20 to 3.4.2 at least) a config map with galaxy.yml values that break because the point to the expected CVMFS path. It used to be I think that giving a galaxy.yml file in the configs of helm used to replace the galaxy.yml in the config map, but now apparently there is a merge going on. The deployment fails with:

```
No handlers could be found for logger "__main__"
Traceback (most recent call last):
  File "/galaxy/server/scripts/galaxy-main", line 299, in <module>
    main()
  File "/galaxy/server/scripts/galaxy-main", line 295, in main
    app_loop(args, log)
  File "/galaxy/server/scripts/galaxy-main", line 142, in app_loop
    attach_to_pools=args.attach_to_pool,
  File "/galaxy/server/scripts/galaxy-main", line 108, in load_galaxy_app
    **kwds
  File "/galaxy/server/lib/galaxy/app.py", line 115, in __init__
    self._configure_tool_data_tables(from_shed_config=False)
  File "/galaxy/server/lib/galaxy/config/__init__.py", line 1077, in _configure_tool_data_tables
    config_filename=self.config.tool_data_table_config_path)
  File "/galaxy/server/lib/galaxy/tools/data/__init__.py", line 80, in __init__
    self.load_from_config_file(single_config_filename, self.tool_data_path, from_shed_config=False)
  File "/galaxy/server/lib/galaxy/tools/data/__init__.py", line 117, in load_from_config_file
    tree = util.parse_xml(filename)
  File "/galaxy/server/lib/galaxy/util/__init__.py", line 236, in parse_xml
    root = tree.parse(fname, parser=ElementTree.XMLParser(target=DoctypeSafeCallbackTarget()))
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 647, in parse
    source = open(source, "rb")
IOError: [Errno 2] No such file or directory: '/cvmfs/main.galaxyproject.org/config/shed_tool_data_table_conf.xml'
```

The only way to overcome this, without changing the chart version, is to make sure that the provided `galaxy.yml` overrides the values injected with CVMFS paths. I had to add the following to my local helm config:

```
configs:
  galaxy.yml:
    ...
    builds_file_path: "{{.Values.persistence.mountPath}}/tool-data/shared/ucsc/builds.txt"
[...]

PR #185, opened 2020-11-10
Consolidated values

almahmoud:
Addresses https://github.com/galaxyproject/galaxy-helm/issues/174 more permanently

This set of values works with both CVMFS enabled and disabled. CVMFS-related configs were wrapped to only take effect if CVMFS is enabled, eliminating the second values file and making a consolidated default that works both with and without CVMFS.

TLDR:
- Removed `helm_package` script. Can now package with normal `helm package`, and use the same values file for CVMFS-enabled and default launches
- `.Values.cvmfs.enabled` affects the default configs now (i.e. setting to enabled and disabled works by default)
- Added `.Values.cvmfs.deploy` to control the CVMFS-CSI conditional dependency (i.e. setting `deploy=true` and `enabled=true` deploys Galaxy with CVMFS. Setting `deploy=false` and `enabled=true` deploys Galaxy with CVMFS configs but without CVMFS chart (i.e. expects CVMFS storage classes to already exist).

Con:
- It might be harder to read the un-templated configs in CM


almahmoud:
@pcm32 Can you verify if this works well for you (just to have it verified with a different set of values than we usually use in GVL/Anvil)?
Also, @nuwang @afgane @pcm32 thoughts on the CVMFS conditional dependency?

This is a fairly consequential change that we have to accommodate for upstream, but should get us back to be more closely following helm standards both in terms of not having out-of-band dependencies (requiring CVMFS storage classes with no way of deploying CVMFS along with the chart) and the consolidation into a single values file.

galaxy/values.yaml
```diff
@@ -230,10 +248,11 @@ configs:
               <param id="k8s_fs_group_id">101</param>
               <param id="k8s_supplemental_group_id">101</param>
               <param id="k8s_pull_policy">IfNotPresent</param>
-              <param id="k8s_cleanup_job">always</param>
+              <param id="k8s_cleanup_job">onsuccess</param>
```
afgane, line 260: Thank you!

galaxy/values.yaml
```diff
@@ -275,10 +294,18 @@ configs:
...
+      tool_config_file: "/galaxy/server/config/tool_conf.xml{{if .Values.cvmfs.enabled}},{{.Values.cvmfs.main.mountPath}}/config/shed_tool_conf.xml{{end}}"
       shed_tool_config_file: "/galaxy/server/config/mutable/editable_shed_tool_conf.xml"
+      tool_data_table_config_path: |-
+        {{ if .Values.cvmfs.enabled -}}
+        {{ .Values.cvmfs.main.mountPath }}/config/shed_tool_data_table_conf.xml,{{.Values.cvmfs.data.mountPath}}/managed/location/tool_data_table_conf.xml,{{.Values.cvmfs.data.mountPath}}/byhand/location/tool_data_table_conf.xml
+        {{- end }}
       tool_dependency_dir: "{{.Values.persistence.mountPath}}/deps"
       job_config_file: "/galaxy/server/config/job_conf.xml"
+      builds_file_path: |-
+        {{ if .Values.cvmfs.enabled -}}
```
afgane, line 315: Shouldn't this conditional be line above? ie, if not enabled, don't include `builds_file_path`.

galaxy/values.yaml
```diff
@@ -288,6 +315,10 @@ configs:
       tool_path: "{{.Values.persistence.mountPath}}/tools"
       tool_data_path: "{{.Values.persistence.mountPath}}/tool-data"
       conda_auto_init: false
+      len_file_path: |-
+        {{ if .Values.cvmfs.enabled -}}
```
afgane, line 328: Same here as with `builds_file_path`

galaxy/values.yaml
```diff
@@ -349,11 +380,50 @@ configs:
...
+        <tool file="data_source/mousemine.xml" />
+        <tool file="data_source/ratmine.xml" />
+        <tool file="data_source/yeastmine.xml" />
+        <tool file="data_source/worm_modencode.xml" />
+        <tool file="data_source/wormbase.xml" />
+        <tool file="data_source/zebrafishmine.xml" />
+        <tool file="data_source/eupathdb.xml" />
+        <tool file="data_source/hbvar.xml" />
+        <tool file="genomespace/genomespace_importer.xml" />
+        <tool file="genomespace/genomespace_push.xml" />
```
afgane, line 417: Do all these work? I think it's pretty safe to omit genomespace entries at least.

galaxy/values.yaml
```diff
@@ -349,11 +380,50 @@ configs:
...
+      </section>
+      <section id="send_data" name="Send Data">
+        <tool file="genomespace/genomespace_exporter.xml" />
+      </section>
```
afgane, line 421: This we can remove.

galaxy/values.yaml
almahmoud: I don't think it can be the line above if we want to keep it valid yaml and not turn the whole thing into a multi-line string. If CVMFS is disabled, it will end up being an empty string which it seems galaxy just ignores and works as if it's not set.

galaxy/values.yaml
almahmoud: I'm not sure, this is from the previous `values-cvmfs`. It doesn't break startup for either case but I have no idea if they work in either.

almahmoud: I think this should probably be a separate file anyway imo

galaxy/values.yaml
almahmoud: Done in https://github.com/galaxyproject/galaxy-helm/pull/185/commits/ce1327a65a2724a02387435719d1ff76ba5434b7

afgane (approved):

Issue #214, opened 2021-01-26
User deployment not working


viktoriaas:
Hi, 
I'm trying to deploy this helm chart under user and it crashes fataly. 
I've already managed to deploy it as admin which works fine but I would like to let users also deply this chart. I use RKE provider, Kuberentes version 1.19.3 and Docker 20.10.2.

Firtly, it is importnant to answer whether this should be possible or if you have thought about it. If yes, following notes are for you. If no, may I ask why? It would certainly help multitenant environments where users might want to run their own instances of galaxy.

1. users are forbidden from creating rbacs therefore I set `rbac:enable: false`. After this, I can deploy the chart.
2. the deployment crashes on postgres StatefulSet. I use some storage provider which doesn't support dynamic provisioning. Therefore I have to create storage separately. I am allowed to use already existing claim for galaxy but not for postgres. And when I create it, stateful set keeps on crashing because it has to have certain naming (it's not clear to me how this naming is determined)
2.1 When I fulfill the naming, the claim is created under root with `drwx------`  permissions. Later, the container itself crashes on `mkdir: cannot create directory ‘/bitnami/postgresql/data’: Permission denied`. I can manually chmod created share (it's on NFS) but it can't work like that. The fix should be somewhere in mounting the claim but there is no trace of postgres configs in chart
3. When storages are fixed, stateful set is running but other components are in Error state after some time.
3.1 galaxy-job + galaxy web + galaxy workflow all fail on 
`sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) FATAL:  password authentication failed for user "galaxydbuser"  `. I haven't manipulated with this value at all.
3.2  I thought it might be connected to that disabled rbac so I created cluster role and binding manually and deplyed into cluster. The helm chart then get deployed, but problem with password persists

Could you advise me if there is **any** way how could I achieve user-deploy possibility?

almahmoud:
Hey! First of all, I'd want to clarify what you mean by the user/admin distinction. My current understanding is that you're referring to something like a Rancher/Kubernetes admin vs namespace-restricted user (but if I misunderstood, this answer might be less relevant than I thought). If you could give us a bit more info to understand more fully the distinction in terms of permissions and specific requirements in your setup.
[...]

PR #215, opened 2021-01-28
Make pod priority class configurable

nuwang:
Fixes issue raised in: https://github.com/galaxyproject/galaxy-helm/issues/214

### DIFF
--- a/galaxy/templates/_helpers.tpl
+++ b/galaxy/templates/_helpers.tpl
@@ -138,6 +138,17 @@ Define pod env vars
 {{- end -}}
 
 {{/*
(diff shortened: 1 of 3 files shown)


afgane (approved):
Why not enable the priority class by default? That was the (implicit) behavior thus far.

Not necessarily related to this PR, but the `_helpers` file has no consistency wrt to naming the defined blocks. Some use camel case, some dashes, and some underscores...

nuwang:
@afgane Thanks for catching that! Made it false while testing and forgot to flip it back. Fixed now.

PR #218, opened 2021-02-08
205 ingress template

### LINKED ISSUE #205: Migrating the ingress template to networking.k8s.io/v1
Related to https://github.com/galaxyproject/galaxy-helm/issues/130

The ingress template uses `extensions/v1beta1` which is deprecated in Kubernetes >=v1.16. You would need to switch to `networking.k8s.io/v1` which is available since 1.8.
You would need to:
- change to `networking.k8s.io/v1` here https://github.com/galaxyproject/galaxy-helm/blob/d12395f121e2ff739b6ad74af54c4facb95ef582/galaxy/templates/ingress.yaml#L5
- add a `pathType: Prefix` and split `serviceName`and `servicePort` into `service.name` and  `service.port.name` here https://github.com/galaxyproject/galaxy-helm/blob/d12395f121e2ff739b6ad74af54c4facb95ef582/galaxy/templates/ingress.yaml#L33

See there for an example: https://kubernetes.io/docs/concepts/services-networking/ingress/

ksuderman:
Update apiVersion for priorityclass-job.yaml and ingress.yaml

Both files used beta versions (v1beta1) in the `apiVersion` field; these have been updated to `scheduling.k8s.io/v1` and `networking.k8s.io/v1` respectively.  This PR also includes minor syntax updates for the `http.paths.path` block, in particular the changes to the service port name and number.

Fixes #205

### DIFF
--- a/galaxy/templates/ingress.yaml
+++ b/galaxy/templates/ingress.yaml
@@ -2,7 +2,7 @@
 {{- $fullName := include "galaxy.fullname" . -}}
 {{- $ingressPath := .Values.ingress.path -}}
 {{- $servicePort := .Values.service.port -}}
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: {{ $fullName }}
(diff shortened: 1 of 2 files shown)


nuwang (approved):
Thanks @ksuderman, this looks good!

PR #221, opened 2021-02-15
Better handling for extraFileMappings

almahmoud:
Encountered as part of the CI setup with `pykube_util.py`

### DIFF
--- a/galaxy/templates/deployment-job.yaml
+++ b/galaxy/templates/deployment-job.yaml
@@ -137,7 +137,7 @@ spec:
               subPath: probedb.py
             {{- range $key, $entry := $.Values.extraFileMappings -}}
             {{- if $entry.applyToJob }}
-            - name: {{ include "galaxy.makeDnsCompliant" (printf "extra-%s-%s" (include "galaxy.getFilenameFromPath" $key) ($key | sha256sum)) }}
+            - name: {{ include "galaxy.getExtraFilesUniqueName" $key }}
               mountPath: {{ $key }}
               subPath: {{ include "galaxy.getFilenameFromPath" $key }}
             {{- end }}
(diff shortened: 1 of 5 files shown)


galaxy/templates/deployment-job.yaml
```diff
@@ -137,7 +137,7 @@ spec:
               subPath: probedb.py
             {{- range $key, $entry := $.Values.extraFileMappings -}}
             {{- if $entry.applyToJob }}
-            - name: extra-files-{{ include "galaxy.getFilenameFromPath" $key | replace "." "-" }}
+            - name: extra-files-{{ include "galaxy.getFilenameFromPath" $key | replace "." "-" | replace "_" "-" }}
```
nuwang, line 140: I think this should be moved to the helper template itself, or have some kind of utility function like `make_k8s_label_compatible` or something?

almahmoud: Yes! i'll just call it `make_dns_compliant` for now

almahmoud: This is all done now, and handled more general cases and the corner case where two files have the same filename but different paths.

almahmoud:
This is ready to be merged imo

galaxy/templates/deployment-job.yaml
```diff
@@ -137,7 +137,7 @@ spec:
               subPath: probedb.py
             {{- range $key, $entry := $.Values.extraFileMappings -}}
             {{- if $entry.applyToJob }}
-            - name: extra-files-{{ include "galaxy.getFilenameFromPath" $key | replace "." "-" }}
+            - name: {{ include "galaxy.makeDnsCompliant" (printf "%extra-%s-%s" (include "galaxy.getFilenameFromPath" $key) ($key | sha256sum)) }}
```
nuwang, line 140: Since this code segment for figuring out the filename has been repeated and is also a bit difficult to understand at first glance, can we move it to a template function, and then document what's going on in the template? e.g. Why is the sha256sum needed here etc? (presumably for uniqueness?)

something like: `get_extra_files_entry_name()` and `get_extra_files_secret_name()` or something?

Also, this particular line seems to be incorrect - should be`%s-%s` instead of `%extra-%s-%s'?

galaxy/templates/deployment-job.yaml
```diff
@@ -184,13 +184,13 @@ spec:
           {{- end }}
         {{- range $key, $entry := $.Values.extraFileMappings -}}
         {{- if $entry.applyToJob }}
-        - name: extra-files-{{ include "galaxy.getFilenameFromPath" $key | replace "." "-" }}
+        - name: {{ include "galaxy.makeDnsCompliant" (printf "%extra-%s-%s" (include "galaxy.getFilenameFromPath" $key) ($key | sha256sum)) }}
```
nuwang, line 187: also here and a couple more below.

nuwang (commented):
Glad to see that this can handle a pretty large number of cases now. Have noted a minor issue.

galaxy/templates/deployment-job.yaml
almahmoud: That is definitely a typo, was supposed to be just `extra`. The `sha256sum` is just for uniqueness (for the case where it's the same filename but different paths). Sure i'll put things in a function

Issue #219, opened 2021-02-11
Problems with file uploading and downloading


viktoriaas:
Hi, 
I don't want to continue in #214 because this issue is different. We have a working Galaxy deployment and added our own tool.

**Problem 1**
When cvmfs is disabled, the PVCs related to them isn't created, right. However, they are later requested when uploading file to Galaxy. In the end, they are not used, the files are stored under _objects_ directory in `galaxy-pvc`. I don't understand why in [values.yaml job.xml](https://github.com/galaxyproject/galaxy-helm/blob/e9167da89481c6ba3db39bb75d524a315f6dae09/galaxy/values.yaml#L217) has both cvmfs PVCs specified, even when disabled. Maybe if these two were included only when cvmfs in **enabled** then the upload wouldn't fail on missing cvmfs pvc.

**Problem 2**
I've manually created cvmfs PVCs, then upload ended okay. When I wanted to view the results, 2 fastq files could be viewed and one txt couldn't. No result could be downloaded, outcome was `404 Not found`.

Here you can see that 2 subsequent queries for preview are different - one HTTP 200, another 404
```
192.168.150.245 - - [11/Feb/2021:14:48:23 +0000] "GET /datasets/9a55ed48fb2654dc/display/?preview=True HTTP/1.1" 200 74301 "https://el-galaxy.cerit-sc.cz/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0.2 Safari/605.1.15"
192.168.150.245 - - [11/Feb/2021:14:48:24 +0000] "GET /datasets/6260c2f98abaa914/display/?preview=True HTTP/1.1" 404 169 "https://el-galaxy.cerit-sc.cz/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0.2 Safari/605.1.15"
```

The 404 preview  save icon led to
```
https://XXX.cz/galaxy/datasets/133c49ca34a2f96a/display?to_ext=txt
```
but actually ended on
```
https://XXX.cz/galaxy/_x_accel_redirect/
```
which displayed 404.
[...]

PR #222, opened 2021-02-16
Fix cvmfs PVC dependency and nginx conf

viktoriaas:
Regarding #219 , I added check into `values.yaml` whether cvmfs is enabled. If true, only then include cvmfs PVCs into `job_conf.xml`

Secondly, I added trailing slash into `configmap-nginx`. Without it, the redirect didn't work and also  [official documentation](https://docs.galaxyproject.org/en/master/admin/nginx.html#sending-files-with-nginx) states that in the nginx, the trailing slash should be present.

### DIFF
--- a/galaxy/templates/configmap-nginx.yaml
+++ b/galaxy/templates/configmap-nginx.yaml
@@ -55,7 +55,7 @@ data:
                 include uwsgi_params;
             }
 
-            location {{ template "galaxy.add_trailing_slash" .Values.ingress.path }}_x_accel_redirect {
+            location {{ template "galaxy.add_trailing_slash" .Values.ingress.path }}_x_accel_redirect/ {
                 internal;
                 alias /;
                 add_header X-Frame-Options SAMEORIGIN;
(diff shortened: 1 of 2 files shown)


almahmoud (approved):

galaxy/templates/configmap-nginx.yaml
```diff
@@ -55,7 +55,7 @@ data:
                 include uwsgi_params;
             }
 
-            location {{ template "galaxy.add_trailing_slash" .Values.ingress.path }}_x_accel_redirect {
+            location {{ template "galaxy.add_trailing_slash" .Values.ingress.path }}_x_accel_redirect/ {
```
nuwang, line 58: It's very strange that this trailing slash would be needed since usegalaxy.org doesn't seem to be using it, and it does work fine on the GVL, which makes me believe that something else may be going on. Still I can't see any harm in merging this, so will do so now pending more data.

nuwang (approved):
Thanks for the fixes @viktoriaas!

PR #223, opened 2021-02-17
Bad typo in job conf

almahmoud:
Missed small typo from https://github.com/galaxyproject/galaxy-helm/pull/222/files which was causing jobs to not be able to run (not unique mountPaths)


almahmoud:
This should be merged without label to overwrite the previous packaging!

PR #224, opened 2021-02-21
[WIP] Finish updating readme

ksuderman:
There are two outstanding issues:

1. Under the **Data Persistence** section the link to the `nfs-provisioner` points to a deprecated repository.  This should be updated to one of the currently maintained options.
2. I am unable to test any of the code blocks as I am still unable to get the charts working on my machine or Jetstream.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,6 @@ accessibility, reproducibility, and transparency of primarily bioinformatics
 data. This repo contains [Helm charts](https://helm.sh/) for easily deploying
 Galaxy on top of Kubernetes.
 
-
 You may [follow this documentation](https://galaxyproject.org/cloud/k8s/) on
 how to use this Helm chart to deploy Galaxy on various managed kubernetes 
 services (e.g., Amazon EKS and Google GKE).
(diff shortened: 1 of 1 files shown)


nuwang:
@ksuderman What was the error in your local environment? Is it related to cvmfs?

ksuderman:
No, I am see other problems on my local machine:

1. There is a problem with the PVC (galaxy-galaxy-pvc)
    Failed to provision volume with StorageClass "ebs": Only support ReadWriteOnce access mode
1. There is a problem with the keycloak secret:
    Secret "keycloak-client-secret-https://rancher.149.165.170.163.xip.io/v1-saml/keycloak/saml/metadata" is invalid:

I mentioned these issues [on Gitter](https://gitter.im/galaxyproject/wg-deployment?at=601efdd2aa6a6f319d0cb7f1).

Oddly, when I try to run the chart on a Jetstream instance I also get an error from the cron-metrics pod that `influxdb.username` is an undefined variable, which I haven't puzzled out since `influxdb.enabled == false` so cron-metrics shouldn't even be in the picture.  But I can fix that one by simply defining the url, username, and password for influxdb in the values.yaml

I will see if I can figure out the problem with the PVC first.

nuwang:
For the storage class, you may need to set `galaxy.persistence.storageClass: nfs` or similar, instead of ebs, since ebs is ReadWriteOnce.

If the influxdb username is required even when influxdb is disabled, that does sound like an oversight.

almahmoud:
@ksuderman Ahhh sorry for missing this earlier, wish I would've noticed the PR earlier. Re the NFS, that's a known issue and I wouldn't worry about it for now. Nuwan and I have talked a bit about new possible charts and one that might be at least a good stop gap for now, just haven't gotten to testing and documenting it. For the other startup issues, given that the biggest remaining task is updating values, it might be worth just using `helm template` if you ever need to see the manifests or verify that the things are deployable (it will error out if there are any manifest-level issues), but no need to necessarily deploy it live, at least to start.
One thing I would start with is just looking at all the defaults in `values.yaml` and including them in the table in the README. If any values are unclear or unintuitive at what they do, feel free to DM me and I can clarify anything of what they're meant to do...
Pointing to a new NFS and launching the chart to verify defaults etc... is also a very important task, but I'd separate it from the bulk of updating the README so that they're not blocking one other.

nuwang:
@ksuderman Would you mind reopening this PR? It seems that merging Alex's PR automatically closed this.

PR #230, opened 2021-03-01
Added canary ingress for scaling on user activity

nuwang:
Added a canary ingress for detecting user activity. This is useful for autoscaling on activity.


afgane (approved):
Tested this in practice and as soon as a request is made on Galaxy, and as long as requests keep coming, the signal to launch/keep a node is active. Really helps with cluster responsiveness.

PR #236, opened 2021-03-04
Separate workflow from job handlers

### LINKED ISSUE #226: Separate workflow from job handlers in values via `applyToWorkflow`
The original chart was designed for 2 handlers (job and web), then a third workflow handler was later added, but the values have not been updated to be able to customize the workflow handler independently. For the most part, `applyToJob` still applies to both the Job handler and Workflow handler.

Things to do:
- [x] In [values.yaml](https://github.com/galaxyproject/galaxy-helm/blob/master/galaxy/values.yaml#L107) add `applyToWorkflow` wherever `applyToJob` and `applyToWeb` exist (including commented examples). Mimic the defaults to be the same as `applyToJob`
- [x] In [deployment-workflow.yaml](https://github.com/galaxyproject/galaxy-helm/blob/master/galaxy/templates/deployment-workflow.yaml) change all mentions of `applyToJob` to be `applyToWorkflow`
- [x] Test the values you added. Tip: create a tmp dir (eg: `mkdir my-tmp-dir`) then run `helm template ./galaxy-helm/galaxy --generate-name --output-dir ./my-tmp-dir` to generate templates. You can then compare the workflow template at `./my-tmp-dir/galaxy/templates/deployment-workflow.yaml` running first with `applyToWorkflow` in values being True then False, and check whether the behavior is what would be expected.

ksuderman:
Closes #226 

***Tasks***

- [x] In values.yaml add applyToWorkflow wherever applyToJob and applyToWeb exist (including commented examples). Mimic the defaults to be the same as applyToJob
- [x] In deployment-workflow.yaml change all mentions of applyToJob to be applyToWorkflow
- [x] Test the values you added. Tip: create a tmp dir (eg: mkdir my-tmp-dir) then run helm template ./galaxy-helm/galaxy --generate-name --output-dir ./my-tmp-dir to generate templates. You can then compare the workflow template at ./my-tmp-dir/galaxy/templates/deployment-workflow.yaml running first with applyToWorkflow in values being True then False, and check whether the behavior is what would be expected.


ksuderman:
I ran the `helm template` command twice, once with `applyToWorkflow: true` and once with `applyToWorkflow: false` in the *values.yml* and then did a `git diff` on the generated files:

```
diff --git a/./false/galaxy-helm/galaxy/templates/deployment-workflow.yaml b/./true/galaxy-helm/galaxy/templates/deployment-workflow.yaml
index 68f9a79..8910cda 100644
--- a/./false/galaxy-helm/galaxy/templates/deployment-workflow.yaml
+++ b/./true/galaxy-helm/galaxy/templates/deployment-workflow.yaml
@@ -128,6 +128,9 @@ cp -aruL /galaxy/server/tools /galaxy/server/database/tools | true;']
             - name: extra-files-probe-script
               mountPath: /tmp/probedb.py
               subPath: probedb.py
+            - name: extra-welcomehtml-ee3410714399628f55d8b0fbdbcc0b1ab19c965ad38e8
+              mountPath: /galaxy/server/static/welcome.html
+              subPath: welcome.html
             - name: galaxy-conf-files
               mountPath: /galaxy/server/config/build_sites.yml
               subPath: build_sites.yml
@@ -182,6 +185,9 @@ cp -aruL /galaxy/server/tools /galaxy/server/database/tools | true;']
         - name: galaxy-conf-files
           configMap:
             name: RELEASE-NAME-galaxy-configs
+        - name: extra-welcomehtml-ee3410714399628f55d8b0fbdbcc0b1ab19c965ad38e8
+          configMap:
+            name: RELEASE-NAME-galaxy-extra-welcomehtml-ee3410714399628f55d8b0fbdbcc0b1ab19c965ad38e8
         - name: galaxy-job-rules
           configMap:
             name: RELEASE-NAME-galaxy-job-rules
```

@almahmoud are these the only changes that should be expected?  LGTM...

galaxy/values.yaml
```diff
@@ -512,6 +514,7 @@ extraFileMappings:
     useSecret: false
     applyToJob: false
     applyToWeb: true
+    applyToWorkflow: true
```
almahmoud, line 517: I would default this to `false` since it only matters in the web handler

almahmoud (changes requested):
Other than the quick default change, looks great! Thanks!
I would just drop the readme change to avoid the conflict and should be good for merge.

I'll try to give the CI some love tomorrow and hopefully fix both the bumping bug and the token issue that's preventing forks from working, then I'll merge this!

galaxy/values.yaml
ksuderman: I guess I forgot to change it back after testing.  Set the default back to `false` now.

PR #239, opened 2021-03-05
Update table in Configuration section

### LINKED ISSUE #238: Update Configuration table in the README with new entries in the values.yml file.
Several new entries in the values.yml file need to be documented in the README and descriptions added.

ksuderman:
Adds to entries from the `values.yml` file to the Configuration section of the README.

There is a small Python program in the scripts directory (`scripts/values_table.py`) that parses the descriptions from the current README and inserts those into the table generated from parsing the `values.yml` file.

Closes #238


ksuderman:
I have started to document some of the new entries in the `values.yml` file, but @almahmoud and @nuwang should likely review the table to ensure we want all of those values included.  At the least likely the `extraFileMappings` entry can be compressed.  I am also unclear on the meanings of some of the entries.  E.g. what is the difference between `postgresql.enabled` and `postgresql.persistence.enabled`?

almahmoud:
This looks like a good start!
You're right that some of the nested values are not needed. I would generally not include any of the values where the Key itself is editable (eg: for `annotations`, we don't use values directly below the key, it's all considered a dict passed to the `annotations` value, so the rest is just looping through arbitrary keys, so no need to add those nested keys, it's enough to just add `annotations`) One way of thinking about it is that the chart would still work fine with `--set annotations={}`, so all the keys under are just our defaults for the dict but would have to be removed/changed depending on the ingress controller or whatever other things are reading these annotations. Hope I'm not doing a horrible job of explaining it, but feel free to ask for specific keys too cause some of it can be confusing!

The `postgresql` question is a good one, and a special case. I think in general we don't want to document any values that are passed to dependency charts. `cvmfs` and `postgresql` are the only two right now, and the only values we need to document for those are `cvmfs.enabled` `cvmfs.deploy` and `postgresql.enabled`. The first one is used by us to change configurations based on whether cvmfs should be expected, the last two are the two values for the conditional dependencies (https://github.com/galaxyproject/galaxy-helm/blob/master/galaxy/requirements.yaml). Otherwise all other values nested under those correspond to the values from the original chart we are inheriting, and it'd be a hassle to document them all here and update them as the upstream chart changes, so we can just point to https://github.com/bitnami/charts/tree/master/bitnami/postgresql#parameters if users want to see further options for that chart for example...

Hope this makes sense, happy to clarify anything further, i'm not always the best at explaining things :)

nuwang:
@ksuderman Great idea to write `scripts/values_table.py`, that should come real handy in future to keep this in sync :-)

ksuderman:
Except when the script has bugs that misses entries in the values.yml file! [Fixed].  It also struck me that the  job, web, and workflow handlers share much of the same configuration,  so I have a [suggested alternative layout](https://github.com/ksuderman/galaxy-helm/blob/238-configuration-table/Configuration.md) with the handlers documented separately from the rest of the configuration.  I am not convinced it is an improvement...

Some other questions I have:

1. If `metrics.enabled == true` does that imply `influxdb.enabled == true` as well?  Does it makes sense to enable the influxdb but  not metrics?
1. The `jobHandlers` has a `priorityClass`, can the other handlers have this as well or only `jobHandlers`?

I am sure I will have more questions as I flesh out the descriptions.

almahmoud:
> I have filled in what I can and the configuration table could use a proof-read. I still have a few questions:
Thanks, this looks great!

> 1. What does the `fullnameOverride` override, i.e. the full name of what exactly?
It fully overrides the name of "galaxy.fullname" helper function: https://github.com/galaxyproject/galaxy-helm/blob/master/galaxy/templates/_helpers.tpl#L15 which is used in most naming: https://github.com/galaxyproject/galaxy-helm/blob/master/galaxy/templates/deployment-workflow.yaml#L4
I think the old descriptions are still fairly applicable:
```

nameOverride | Override the name of the chart used to prefix resource names. Defaults to {{.Chart.Name}} (i.e. galaxy)

fullNameOverride | Override the full name used to prefix resource names. Defaults to {{.Release.Name}}-{{.Values.nameOverride}}

```
These are fairly common in the helm world so I think for the most part users should be somewhat familiar: eg:  https://github.com/bitnami/charts/blob/master/bitnami/postgresql/values.yaml#L42 

> 2. If metrics is enabled does the Influxdb have to be enabled? Can metrics work without a database?
>    a. does it make any sense to enable influxdb if metrics are not enabled?
I think not, but @nuwang can better answer

> 3. The `jobHandlers` is that only handler to have `priorityClass` defined, can the other handlers have a `priorityClass`?
Yes it is only for the jobHandler. In fact it is actually not even used in the job handler directly, it's just for the [priority class template ](https://github.com/galaxyproject/galaxy-helm/blob/master/galaxy/templates/priorityclass-job.yaml)(and the priority class is subsequently used by the k8s runner). It might actually make sense put it under `jobs.priorityClass` next to `jobs.rules` since it doesn't really affect the handler directly like the other settings under job handler? @nuwang @afgane opinions/objections?

galaxy/scripts/values_table.py
```diff
@@ -0,0 +1,134 @@
+
```
almahmoud, line 1: I think you should move this under a new dir `galaxy-helm/scripts` not `galaxy-helm/galaxy/scripts` since it's a convenience script for us so no reason to have it be packaged as part of the pushed chart. Removing the other scripts unrelated to the chart from there in https://github.com/galaxyproject/galaxy-helm/pull/251

nuwang:
@ksuderman The PR dropping support for helm2 was just merged, would you mind regenerating the docs? We can merge immediately after?

ksuderman:
@nuwang is that what you had in mind?  I simply added a note that support for Helm 2 has been removed.

nuwang:
@ksuderman Sorry I wasn't clear, what I meant was that this PR was merged: https://github.com/galaxyproject/galaxy-helm/pull/234 and it changes the content of the configuration section again, so rerunning your script to regenerate the table would be good.

ksuderman:
I am closing this PR and will open another.

PR #253, opened 2021-03-13
Copy converters

almahmoud:
Interim solution to solve the problem for the k8s runner of converters that expect scripts but also have requirements and therefore use mulled containers. This should not be an issue with pulsar as the tool directory (and hence scripts) will be staged into the container if I understand correctly.


almahmoud:
Closing in favor of https://github.com/galaxyproject/galaxy-helm/pull/265

PR #254, opened 2021-03-13
Change image to galaxy-min 21.01

nuwang:


almahmoud:
We should probably either merge this or change `appVersion` back.
On a related note, do we also want to revert back some versions and do a big 4.0 release with 21.01? We can start by just forcing it back to `4.0.0` or `4.0.0-dev` after which we don't use labels on PRs for a few days and just overwrite that single version until we are satisfied and we release 4.0.0 and restart bumping from there? Purely aesthetic, so happy to not, but we're still close to when the 4's started being packaged so we could still just do a force cleanup hopefully without affecting anyone

afgane:
Can we just push 4.0 out quickly?

PR #257, opened 2021-03-16
Metrics disabled by default

ksuderman:
Set `metrics.enabled: false` as the default value in the `values.yml`.

Metrics should only be enabled when an influxDB is configured and available, which is not the *default* when installing Galaxy with just the Helm chart.


ksuderman:
I need to clean up my fork first.

PR #271, opened 2021-04-05
Migrations overhaul

almahmoud:
Main changes:

- Moved init containers needing to run once at startup to movable helper template
- Added value: `initContainers.mainHandler="web" # valid values are "web" "job" or "workflow"` which decides which handler should have the DB initialization init containers
- Added value `initContainers.createDatabase=true` which will create the DB on the first run and also run migrations at each startup (to ensure that upgrading the image is still enough to perform an upgrade). If set to false, the DB will be expected to have been initialized and only migrations will be run
- Only the main container will wait on the DB directly, the rest of the containers wait for the DB initialization signal in the shared filesystem.
- ~All handler pods don't have a waiting init container anymore, they simply wait in the main container. This is slightly faster than waiting in init containers, but we can still move them to init containers if we don't mind the extra few seconds and want main container probes to be cleaner~
All handlers have an init container waiting for the signal on the shared filesystem, and only the main container does the init mounts and init DB. This seems cleaner than having a wait in the main container and a few seconds are really not that important to save right now. I made this init container use the galaxy image to keep it pulled as early as possible
- There still seems to be a race condition caused by multiple handlers coming up at once (the DB would get locked and become unresponsive when job and workflow handler would come up exactly at the same time), but as of 21.01 the `create_db.sh` was solved in https://github.com/galaxyproject/galaxy/pull/11753 so we can now bypass migrations. It's not an ideal solution, but to avoid the race condition as a stop gap, I just scattered the startups by 10 seconds. After the DB init signal is given, web handler will start right away, job handler will wait 10 seconds, and workflow handler will wait 20 seconds. This is enough to seemingly reliably avoid the race condition while keeping overhead minimal. This should only be temporary though.

What this accomplishes:
- No more handler restarts
- DB created much faster when migrations are not needed
- DB init containers can be moved to any handler


galaxy/values.yaml
```diff
@@ -118,16 +121,20 @@ persistence:
...
 #     persistentVolumeClaim:
 #       claimName: shared-data-pvc
 
-# extraVolumeMounts:
+extraVolumeMounts: []
 #   - name: shared-data
 #     mountPath: /mnt/project/shared-data
 
-# extraInitContainers:
+initContainers:
```
nuwang, line 133: Do we need to make this user configurable? My intuition here is that we don't and can just do this in the web handler always? Otherwise LGTM

almahmoud: I don't mind making it not configurable, main reason was to be able to spin up a specific handler without requiring the other ones, and move it in case one wants to start with multiple replicas (since ideally the init container should be only once). I'm now thinking maybe going back to making it jobs is the way to go, that way it's independent of the handlers altogether?

almahmoud:
Updated version highlights:
- DB creation/migrations, and init mounts (copying config/galaxy-lib tools etc...) are now run in a k8s job independent of any handler
- Handlers have init containers waiting for both the db to be initialized and the init mounts to be done
- Adds `*handlers.startupDelay` values, used to slightly scatter handlers
- Main accomplishments:
  - Eliminates the crashbackloop on handlers
  - Creates DB from scratch on first startup (faster), and only runs migrations on upgrades

Considerations:
1) Why job vs init container?
Having the migrations happen independently from handlers seems both conceptually better and is a more extensible implementation. Main benefit that I thought of for using jobs vs init containers:
  - Any handler can now scale without duplicating migration scripts
  - This allows to easily "pre-create" Galaxy instances that can be simply spun up. Essentially running with `replicaCount=0` on all handlers is an out-of-the box way of preparing the database and shared filesystem for an upcoming Galaxy that can then just be scaled up

2) Why not use helm hooks?
TLDR: `pre-` hooks don't work because the init db script needs postgres to be deployed, `post-` hooks don't work when using `--wait` because they don't come up until the command returns, but the command can't return until handlers are up which wait for the hook jobs, hence a recursive wait (and we need `--wait` for Leo in AnVIL).
Solution: Init job and lock filenames per Helm release, but deployed as normal jobs

galaxy/templates/deployment-job.yaml
```diff
@@ -41,48 +41,14 @@ spec:
...
-          volumeMounts:
-            - name: galaxy-data
-              mountPath: {{ $.Values.persistence.mountPath }}
-        - name: {{ $.Chart.Name }}-init-mounts
-          image: "{{ $.Values.image.repository }}:{{ $.Values.image.tag | default $.Chart.AppVersion }}"
+        - name: {{ $.Chart.Name }}-wait-db
+          image: "{{ $.Values.image.repository }}:{{ $.Values.image.tag }}"
           imagePullPolicy: {{ $.Values.image.pullPolicy }}
-          command: ['sh', '-c', {{ include "galaxy.init-container-commands" $ | squote }}]
+          command: ['sh', '-c', '{{ include "galaxy.init-container-wait-command" $ }} sleep {{ $.Values.jobHandlers.startupDelay }}; sleep {{ $.Values.jobHandlers.startupDelay }};']
```
afgane, line 47: Why 2 `sleep` commands?

galaxy/templates/job-init-db.yaml
```diff
@@ -0,0 +1,159 @@
...
+          command:
+            - 'sh'
+            - '-c'
+            - >
+              if (ls /galaxy/server/config/mutable/ | grep -q "db_init_done*");
+                then /galaxy/server/manage_db.sh upgrade && echo "Done" > /galaxy/server/config/mutable/db_init_done_{{.Release.Revision}};
+                else (/galaxy/server/create_db.sh && echo "Done" > /galaxy/server/config/mutable/db_init_done_{{.Release.Revision}};);
+              fi;
+          {{- else }}
+          command:
```
afgane, line 41: Second `command`?

almahmoud: One for `initJob.createDatabase=true` which expects to create the database on the first run and find signals from the create on the NFS, and another for `initJob.createDatabase=false` which always just runs migrations. The option only matters on startup, after startup they will both just run migrations, the second one would just know not to wait to create on the first run

galaxy/templates/deployment-job.yaml
almahmoud: It shouldn't be oopsie, thanks for catching will fix now

afgane (approved):

galaxy/templates/job-init-db.yaml
```diff
@@ -0,0 +1,159 @@
...
+      annotations:
+        checksum/galaxy_conf: {{ include (print $.Template.BasePath "/configs-galaxy.yaml") . | sha256sum }}
+        checksum/galaxy_rules: {{ include (print $.Template.BasePath "/configmap-galaxy-rules.yaml") . | sha256sum }}
+        checksum/galaxy_extras: {{ include (print $.Template.BasePath "/configmap-extra-files.yaml") . | sha256sum }}
+    spec:
+      restartPolicy: OnFailure
+      initContainers:
+        - name: {{ .Chart.Name }}-init-postgres
+          image: alpine:3.7
+          command: ['sh', '-c', 'chown 101:101 {{ .Values.persistence.mountPath }}; until nc -z -w3 {{ template "galaxy-postgresql.fullname" . }} 5432; do echo waiting for galaxy-postgres service; sleep 1; done;']
```
nuwang, line 23: Setting the FsGroup should do the same thing as the chown here right?

galaxy/templates/job-init-db.yaml
```diff
@@ -0,0 +1,159 @@
...
+          volumeMounts: 
+            - name: galaxy-data
+              mountPath: /galaxy/server/database
```
nuwang, line 26: I have this recollection that there's a variable for the mountPath?

almahmoud: Yep you're right, thanks for the catch! Un-hardcoded in the new PR in https://github.com/galaxyproject/galaxy-helm/commit/0404a3a4977328f9e86a41afba1425c9d3dd8110

galaxy/templates/job-init-db.yaml
almahmoud: Not entirely sure, we've always done the chown, so I assumed it was needed, but I haven't tried without. In general, it doesn't add extra startup time, so my guess for now is that there's nothing lost by keeping it

PR #274, opened 2021-04-05
Change default to deploy postgres operator

almahmoud:
Continuing the idea that `helm galaxy .` should work out-of-the-box, I think we should be vocal about recommending to install postgres-operator globally and set it to `false` in our production situations, but keep it deployed by default so that a newcomer can get a running simple galaxy with the default values

### DIFF
--- a/galaxy/values.yaml
+++ b/galaxy/values.yaml
@@ -191,9 +191,9 @@ affinity: {}
 
 postgresql:
   enabled: true
-  # Whether to deploy the postgresl operator. In general, we recommend installing the operator
-  # globally.
-  deploy: false
+  # Whether to deploy the postgresl operator.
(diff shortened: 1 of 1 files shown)


afgane:
Does this need to be befriended with a config change in cloudman-boot or CloudLaunch app registry?

almahmoud:
Yes-ish. If we want to use 4.0 without changing other things (but not good except for only running one galaxy) we have to leave it true for now (You can also theoretically use it with multiple deployments but currently testing in specific scenario with nodeselector/namespace isolation separating the operators which so far seems to not make them interfere with one another)
For production version we need https://github.com/galaxyproject/cloudlaunch-registry/pull/35 and a friend in cloudman-boot adding a task installing the operator

afgane (approved):

PR #277, opened 2021-04-08
CVMFS Cloud

almahmoud:
Needs small extra change in `_helpers.tpl` with https://github.com/galaxyproject/galaxy-helm/pull/271

Friends with https://github.com/CloudVE/galaxy-cvmfs-csi-helm/pull/11

### DIFF
--- a/galaxy/values.yaml
+++ b/galaxy/values.yaml
@@ -166,7 +166,7 @@ ingress:
   hosts:
     - host: ~
       paths:
-        - path: "/"
+        - path: "/galaxy"
   tls: []
   #  - secretName: chart-example-tls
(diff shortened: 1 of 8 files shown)


galaxy/values.yaml
```diff
@@ -166,7 +166,7 @@ ingress:
   hosts:
     - host: ~
       paths:
-        - path: "/"
```
afgane, line 169: Why this change?

almahmoud: Because on line 165 above it's always been set to default to `/galaxy` and the two paths should not mismatch.

afgane (approved):

PR #282, opened 2021-04-20
[21.05] Add changes needed for ITs

almahmoud:
All permissions on ingresses and services (need at least create/delete/get)


galaxy/templates/rbac-job.yaml
```diff
@@ -10,8 +10,11 @@ rules:
 - apiGroups: [""]
   resources: ["pods", "pods/log"]
   verbs: ["get", "list", "watch"]
-- apiGroups: ["batch", "extensions"]
-  resources: ["jobs"]
+- apiGroups: ["batch", "extensions", "networking.k8s.io"]
+  resources: ["jobs", "ingresses"]
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```
nuwang, line 15: Do we need to update/patch ingresses? I was thinking, create and delete only?

almahmoud: Do we need update/patch for jobs if you know? (I don't think we do?) If not, i'm pro removing them altogether

almahmoud: Actually it might be used to scale down a job now that I think about it. I'll just remove it for services and ingresses for now

almahmoud: Done in https://github.com/galaxyproject/galaxy-helm/pull/282/commits/50ecf41874b5929a4a9e5b3b894ade752db443d8

almahmoud:
Since ITs will be in 21.05, I can make this PR a general IT change for that

galaxy/values.yaml
```diff
@@ -261,8 +261,8 @@ configs:
           {{- end }}
         k8s_namespace: "{{ .Release.Namespace }}"
         k8s_galaxy_instance_id: "{{ .Release.Name }}"
-        k8s_run_as_user_id: "101"
-        k8s_run_as_group_id: "101"
+        k8s_run_as_user_id: "0"
+        k8s_run_as_group_id: "0"
```
nuwang, line 265: Does this need to be run as root?

almahmoud: No, removing them to nullify is better, as long as it doesn't affect other tools negatively

PR #284, opened 2021-04-21
Add k3s tests

almahmoud:
3 tests per push:
- Linting (https://github.com/almahmoud/galaxy-helm/actions/runs/769508288)
- Deploy on k3s with CVMFS (https://github.com/almahmoud/galaxy-helm/runs/2397202681)
- Deploy on k3s without CVMFS (https://github.com/almahmoud/galaxy-helm/actions/runs/769508286)

For the latter two, all the probes passing signals a successful deployment.


afgane (approved):
This is awesome!

nuwang:
I think this is great. One improvement perhaps is that the test shouldn't pass unless Galaxy is responding perhaps? Right now, manual inspection seems to be necessary?

almahmoud:
It only passes right now if all the handlers are marked healthy, which means the GET probe worked on the web handler and the heartbeats are working on the other handlers. There is no functional test for Galaxy running jobs, but it is testing that Galaxy can respond and can communicate with the database, which I think is a good relatively fast indication of whether a PR is breaking the chart deployment. There is no external testing for the ingress either I suppose.

PR #287, opened 2021-04-26
Removing dependency on CVMFS for getting tools

almahmoud:
Main improvements:
- Can run Galaxy with containerized tools pre-installed (mimicking `cloud.galaxyproject.org` repo) even with CVMFS disabled
- Startup does not rely on CVMFS caching, and the timing is a lot more reliable. In my experience, the untaring strictly needed for startup happens before the database is ready, thus eliminating any wait as Galaxy will have all startup-needed config files on disk already by the time it receives the signal to startup
- Adds support for using the init jobs to download and untar tools repo from bucket archives, but retains support for using CVMFS:
  - Values changed to allow for dynamically adding any number of CVMFS PVCs. See `.Values.cvmfs.galaxyPersistentVolumeClaims`.
  - Friends with https://github.com/CloudVE/galaxy-cvmfs-csi-helm/commit/70c1806ad3d9299c4b13a35c6027bd65c79a6532#diff-4bc5a34cd72923c34ffd1a0259a110551ab699798aec095baa52348e79b83901R101 and https://github.com/galaxyproject/galaxy/pull/11899


almahmoud:
For 21.01 or below will need `--set initJob.externalConfs.enabled=false` and the commented portions under cvmfs enabled 
```
cvmfs:
  repositories:
    cvmfs-gxy-data: data.galaxyproject.org
    # cvmfs-gxy-cloud: cloud.galaxyproject.org
  enabled: false # create the PVCs
  deploy: false # deploy the cvmfs-csi chart
  galaxyPersistentVolumeClaims:
    data:
      storage: 1Gi
      storageClassName: cvmfs-gxy-data
      mountPath: /cvmfs/data.galaxyproject.org
    # cloud:
    #   storage: 1Gi
    #   storageClassName: cvmfs-gxy-cloud
    #   mountPath: /cvmfs/cloud.galaxyproject.org
```

Backwards incompatible change:
`cvmfs.(data | main).pvc.storage --> cvmfs.galaxyPersistentVolumeClaims.[]`
which allows to dynamically add any number of repositories.
In CVMFS chart all repos are disabled by default now, so need to add
`helm install galaxy-cvmfs-csi --set repositories.cvmfs-gxy-data=data.galaxyproject.org`

galaxy/values.yaml
```diff
@@ -132,6 +132,24 @@ extraVolumeMounts: []
 
 initJob:
   createDatabase: true
+  externalConfs:
```
nuwang, line 135: I feel like this should be renamed to something like 'downloadToolConfs' or similar, as it could be confused with configuration in general?

galaxy/templates/jobs-init.yaml
```diff
@@ -107,14 +180,38 @@ spec:
...
+            - name: cvmfs-gxy-{{ $key }}
+              mountPath: {{ $entry.mountPath }}
+            {{- end -}}
             {{- end }}
             {{- if .Values.extraVolumeMounts }}
             {{- .Values.extraVolumeMounts | toYaml | nindent 12 }}
             {{- end }}
+        - name: {{ .Chart.Name }}-init-cloud-repo
+          image: alpine:3.7
+          command: ['sh', '-c', 'wget -O startup.tar.gz {{ .Values.initJob.externalConfs.archives.startup }} && tar -xvf startup.tar.gz > /dev/null && echo "Done" > /galaxy/server/config/mutable/init_clone_done_{{.Release.Revision}}']
```
nuwang, line 193: It seems like all archives are downloaded regardless of the setting? Should this be conditional instead?

galaxy/values.yaml
```diff
@@ -132,6 +132,24 @@ extraVolumeMounts: []
...
+    enabled: true
+    # Uses Galaxy's shared file system volume
+    volume:
+      subPath: cvmfsclone # on galaxy-data volume (avoid config, tools, lib, etc...)
+      mountPath: /cvmfs/cloud.galaxyproject.org
+    archives:
+      # A tar.gz publicly accessible archive containing AT LEAST conf files and XML tool wrappers.
+      # Meant to be enough for Galaxy handlers to startup
+      startup: https://storage.googleapis.com/cloud-cvmfs/startup.tar.gz
```
nuwang, line 144: Should we simplify this by accepting a url to the archive instead (instead of having 3 settings named startup, running and full)?

galaxy/templates/jobs-init.yaml
almahmoud: No. We need all of them. The startup one needs to contain at least just the XMLs which is enough for the handlers to load <2Mb). The second one needs to contain at least all wrappers and scripts, and the last one can container anything (mostly adding tool test data).

galaxy/values.yaml
almahmoud: I liked the three so that we are able to optimize startup, but leave the option for someone else to just tar up all their confs and use a single full tar everywhere. So in the view of the chart there are always three different stages of decompressing, but in reality someone could just put the full one for all three settings, hence the comment that each tar needs to containt _at least_ a certain potion.

afgane (approved):

PR #294, opened 2021-05-14
Whitespace `k8s_persistent_volume_claims`

almahmoud:
Better fix would be to handle this as a proper YAML list in the k8s runner, and expect it to be json/yaml text in XML (as are the other configs). I can PR that to the runner and this won't be necessary anymore


afgane (approved):

almahmoud:
Merging this cause it doesn't hurt even if adding fix to the runner as well

PR #297, opened 2021-06-04
Load extraFileMappings using .Files.Get if a path is defined.

ksuderman:
A small modification to the `configmap-extra-files.yaml` template to load the file content with `.Files.Get` if a `path` entry has been defined.  For example:

```
extraFileMappings:
  /galaxy/server/lib/galaxy/objectstore/irods.py:
    applyToWeb: true
    path: files/lib/my_code.py
```


almahmoud:
Made https://github.com/galaxyproject/galaxy-helm/pull/298 to make the CI run from fork PRs and rebased over that. Once tests pass, this LGTM

almahmoud (approved):
Assuming you've done at least one manual testing to make sure the indentation is all fine, this looks good to me!

almahmoud:
@ksuderman Merged, but thought right after: could you please follow-up with a README PR showing the option?

almahmoud:
Something I noticed today with the dev env, this makes upgrades not portable, i.e. can't upgrade from a different machine unless all the files added with `path` have been copied as well, even if doing `--reuse-values` (since the value is reused, but the manifest is regenerated and `.Files.Get` is re-evaluated on each upgrade. In light of that, my vote is that we should remove `.path`, and discourage it. Alternatively, we can keep it as a dev-only option, an add an actual helm warning that would print at `helm install` about the feature and effects on upgrades when used.
To accomplish the same thing without the extra value, one could add `extraFileMappings."/my/file/".contents="{{ .Files.Get '/path' }}"` and have the same (non-portable) effect. Otherwise, the best practice is probably to stick with `--set-file  extraFileMappings."/my/file/".contents=/local/path` which gets the contents of the local file but makes upgrades portable by adding the contents themselves to the values.

@ksuderman @nuwang @afgane thoughts?

ksuderman:
-1 for removing.  Although we could document the limitations and drawback and we should definitely point out that linking to files outside the chart is for development purposes only and strongly discouraged in production.

However, I am not sure I follow the problem with upgrades. My understanding is that Helm won't copy files from outside the chart with `.Files.Get` for this reason.  So any file included with `path:` will already be in the chart and should be available for any upgrades.  If a user has included (via some mechanism) files not in the chart wouldn't they face the same problem when doing an upgrade from a different machine?

I typically don't use `--set` or `--set-file` when installing with Helm and instead have installation specific value yaml files and including HTML, XML, and Python code directly in the yaml makes them large and unwieldy.

nuwang:
Can you clarify how you're using this at the moment @ksuderman? My understanding was that `Files.Get` only works with files already bundled in the chart. In which case, do we need it to be configurable through values? Can we instead directly edit the config map to include it, since this would be a developer only option?

almahmoud:
I agree with Nuwan that for files bundled with the charts one can just set the entire `.Files.Get` line instead of needing an extra value. Even if you don't want to hardcode in the template, you can still do it in values very easily.

Instead of:
```
extraFileMappings:
    /my/file:
       path: local/path
```
you can change your values to
```
extraFileMappings:
    /my/file:
       contents: "{{ .Files.Get local/path }}"
```
to accomplish the same thing, unless i'm missing something, since the contents are already run through `tpl`.

Given the extra value introduces the possibility for confusion (it took me a bit to understand why upgrades from rancher were failing on the ChatOps instances), and the same thing can be accomplished without it, what is the value of keeping it?

ksuderman:
@nuwang yes, `.File.Get` only works with files in the chart, so the files have to already be included, or (as I have been doing) you have to install from source.  My intended use case is to be able to inject Python code changes into a running Galaxy instance without having to build and deploy Docker images or paste large amounts of Python into a YAML file.

@almahmoud If it is causing problems and `"{{ .Files.Get file }}"` is equivalent then you can remove it.  Although I am confused how it caused upgrades to fail since, as far as I know, I am the only one to use it.

almahmoud:
I switched to using it in the dev env scripts after merging the PR: https://github.com/CloudVE/galaxy-k8s-dev-env/blob/f9c25ff152f44a136923e882c267b91822eef037/update_links.sh#L58 and when I tried to manually upgrade the chatops deployed instance from rancher, the upgrade failed, after which I noticed it’s because all the injected files became empty.
Ultimately, I don’t particularly care if we keep or remove the feature, I’m personally switching back to set-file in my scripts cause i want to be able to make changes from multiple places, but if we keep it we need to print a clear warning when the chart is installed using it, letting users know that if they had added any files in one install and upgrade from repository or another clone later, they’ll lose what they had injected the first time.

ksuderman:
Ah, I did not realize you started using `path:`.  Sorry for the pain.  I'll submit a PR that fails the install/upgrade if the file isn't loaded.  I sort of assumed `.Files.Get` would at least warn if it couldn't load the file.  Silly me.

But, I am still trying to wrap my head around why that upgrade failed for you. When you say:

 >I tried to manually upgrade the chatops deployed instance from rancher, the upgrade failed 

What do you mean by manually upgrade from Rancher?  Do you mean that you edited one of the YAML/Configs in Rancher?   I can make an upgrade fail for lots of reasons, but I can edit the YAML in Rancher and redeploy a pod without losing any injected files.  And as long as the files are available in the chart directory you should be able to install/upgrade from copies or clones of the repo.  If you're upgrading using the packaged chart, i.e: `helm upgrade galaxy cloudve/galaxy ...` then yeah, I don't think using `path:` is an option until Helm supports loading files from outside the chart.

almahmoud:
In my case, the injected files are available in the chart directory only in the clone made as part of the script in the GitHub workflow, symlinked from the Galaxy PR, which would be a pain to have to reproduce for an upgrade. I understand how copy-pasting the files into values manually is annoying, but having the script generate `--set-file` was probably the best, although ugly, since I could upgrade those instances from wherever.
By manually from rancher, I mean from the `cloudman-server` pod. That's generally where I make helm upgrades when I want to change values in a GVL cluster. So I went into the cloudman container and did a `helm upgrade pr-9-galaxy -n pr-9 cloudve/galaxy --reuse-values --set image.repository=` expecting to just keep everything the same and just have a different image repo (to have the base ref image built from the `alex-k8s` org fork instead of `galaxyproject`), but Galaxy wouldn't load up, and then I realized the injected files had become empty when I did the upgrade from the packaged chart. Editing manifests directly (whether in the Rancher UI or `kubectl apply/patch`) still works ofc. Regarding loading from outside the chart, that might actually make it worse imo, since it would make it more likely that a file injected in one install/upgrade would not be at the next upgrade when done from a different machine.
I personally think that ultimately because the functionality only makes sense for specific files packaged in the chart and not general use (especially since for ad-hoc more general use `contents: "{{ .Files.Get local/path }}"` accomplishes the same thing), that it doesn't seem conceptually like the kind of configurable value we want to expose to users explicitly, but in any case a warning would be enough if it's helpful to keep it for whatever reason.

PR #308, opened 2021-06-25
More Nginx Tweaks

almahmoud:


galaxy/values.yaml
```diff
@@ -627,3 +627,13 @@ nginx:
...
+  resources:
+    # We recommend updating these based on the usage levels of the server
+    requests:
+      cpu: 100m
+      memory: 500M
+      ephemeral-storage: 1Gi
+    limits:
+      cpu: 2
+      memory: 3G
+      ephemeral-storage: 10Gi
```
nuwang, line 639: Maybe this should be parameterized to be the same as the client-max-body-size?

almahmoud: Changed defaults to 10/100Gi

PR #318, opened 2021-08-19
Chown NFS again in mounts

almahmoud:
Config files copied onto the NFS were ending up owned by root. This makes everything uniformly owned by 101 as is expected right now

### DIFF
--- a/galaxy/templates/_helpers.tpl
+++ b/galaxy/templates/_helpers.tpl
@@ -151,7 +151,8 @@ cp -anL /galaxy/server/config/data_manager_conf.xml.sample /galaxy/server/config
 cp -anL /galaxy/server/config/tool_data_table_conf.xml.sample /galaxy/server/config/mutable/shed_tool_data_table_conf.xml;
 cp -aruL /galaxy/server/tool-data {{.Values.persistence.mountPath}}/;
 cp -aruL /galaxy/server/tools {{.Values.persistence.mountPath}}/tools | true;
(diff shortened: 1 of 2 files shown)


galaxy/templates/_helpers.tpl
```diff
@@ -151,7 +151,8 @@ cp -anL /galaxy/server/config/data_manager_conf.xml.sample /galaxy/server/config
 cp -anL /galaxy/server/config/tool_data_table_conf.xml.sample /galaxy/server/config/mutable/shed_tool_data_table_conf.xml;
 cp -aruL /galaxy/server/tool-data {{.Values.persistence.mountPath}}/;
 cp -aruL /galaxy/server/tools {{.Values.persistence.mountPath}}/tools | true;
-echo "Done" > /galaxy/server/config/mutable/init_mounts_done_{{.Release.Revision}}
+echo "Done" > /galaxy/server/config/mutable/init_mounts_done_{{.Release.Revision}};
```
nuwang, line 154: Since it's running as 101 now, this chown probably won't work?

almahmoud: I added the 101 to the DB pods so that the "lock" files are not owned by root. The pod that runs this is still running as root: https://github.com/galaxyproject/galaxy-helm/pull/318/files#diff-08a66afc08decb4201f61a1f5a3bb2870a5880ca4eedbf66144df2485485fc4fR155

almahmoud:
Opened https://github.com/galaxyproject/galaxy-helm/issues/320 to track further improvements

PR #323, opened 2021-08-30
Training materials web hook

ksuderman:
The GTN added a [webhook to Galaxy](https://docs.galaxyproject.org/en/master/admin/special_topics/gtn.html) that enables simple navigation from an online tutorial to the tool(s) being used in the tutorial and back again.  This webhook can be enabled in Galaxy with a simple change to the Nginx configuration to proxy `/training-material/` to the GTN training material server.


galaxy/templates/configmap-nginx.yaml
```diff
@@ -61,7 +61,9 @@ data:
                 add_header X-Frame-Options SAMEORIGIN;
                 add_header X-Content-Type-Options nosniff;
             }
-
+            location /training-material/ {
+                proxy_pass https://training.galaxyproject.org/training-material/;
```
almahmoud, line 65: Shouldn’t you use the value here instead of hard coding?

almahmoud (changes requested):

galaxy/templates/configmap-nginx.yaml
```diff
@@ -61,7 +61,9 @@ data:
                 add_header X-Frame-Options SAMEORIGIN;
                 add_header X-Content-Type-Options nosniff;
             }
-
+            location /training-material/ {
```
nuwang, line 64: Should the proxy prefix be added here?
```suggestion
            location {{ template "galaxy.add_trailing_slash" .Values.ingress.path }}training-material/ {
```

ksuderman: No, I don't believe so; I tried that first and it didn't work during my testing.  Even when using a proxy prefix the expected location is `/training-material/` and not `/galaxy/training-material/` (or whatever the proxy prefix is).

galaxy/templates/configmap-nginx.yaml
ksuderman: I was thinking of that, but decided to hard code since the values will not change and are not something that is parametrizable.

galaxy/templates/configmap-nginx.yaml
nuwang: Hmm, so what happens if you install galaxy at a proxy prefix? The ingress wouldn't be intercepting `/` at that point anyway, so would this break?

ksuderman: That "should be" handled by adding an ingress path https://github.com/galaxyproject/galaxy-helm/blob/50e2e6f4bee6a800fa06c21958ffa05ae039faff/galaxy/values.yaml#L197

galaxy/templates/configmap-nginx.yaml
ksuderman: Not sure what I meant when I said this is not parameterizable... the value will likely never change, but it is now taken from `.Values`

galaxy/templates/configmap-nginx.yaml
ksuderman: My last comment was incorrect, but the training hook still works with some caveats.  I suspect there would be problems if you wanted to redirect to different training-material servers for some reason, and when you enable/disable the web hook on one Galaxy instance you enable/disable it on all of them.  So [https://example.com/galaxy2]() can't disable the web hook if [https://example.com/galaxy1]() has in enabled.

nuwang (approved):

PR #326, opened 2021-09-13
Add maintenance cronjob to delete tmp files older than walltime limit

nuwang:


galaxy/templates/cronjob-maintenance.yaml
```diff
@@ -0,0 +1,43 @@
...
+spec:
+  schedule: "0 2 * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          securityContext:
+            runAsUser: 101
+            runAsGroup: 101
+            fsGroup: 101
```
nuwang, line 16: Should be using Galaxy security context.

almahmoud: Done in https://github.com/galaxyproject/galaxy-helm/pull/326/commits/68cfbbcf4102e38be4755e4d0342a1e540e4130d

Issue #330, opened 2021-09-30
SVG files not being displayed in the new user Welcome page


ksuderman:
The root cause of the SVG files not being displayed is they are being served without a `Content-Type` header.  **Note** Firefox *will* display the SVG files, the problem only appears in Safari and Chrome.  It may exist in other browsers, but these are the only three I checked.

The most common cause of errors with the Content-Type header is a missing `mime.types` file.

### Solutions Attempted (Failed)

1. Added `include /etc/nginx/mime.types;` to the `nginx.conf` as suggested on [StackOverflow](https://stackoverflow.com/questions/51045084/nginx-treating-svg-as-plain-text) and [several other](https://serverfault.com/questions/874597/do-i-need-to-configure-svg-images-showing-in-nginx/874600#874600) places.  I also ensured the `mime.types` file does contains an entry for .svg files.

2. UWSGI loads mime types from `/etc/mime.types` but this file does not exist in the web handler pod/container so I mounted the file as a configmap in Kubernetes

Neither of the above resolved the problem.
### Investigate

Interactions between Nginx and UWSGI.

dannon:
@astrovsky01 Is this the same problem you mentioned to me yesterday?

astrovsky01:
Yes

almahmoud:
I have a fix for this that involves adding the static content onto the NFS (which is shared across nginx and the web/uwsgi pods) and then having NGINX serve static content directly. (Right now, the request is sent by NGINX to UWSGI, who has access to the file off the container image, and then passes it back to NGINX. I believe when uWSGI passes the response, it types it as `application/html` as that's what NGINX always gets, even with `mime.types` explicitly included and with default type set to svg)

almahmoud:
I have a fix for this that involves making the static content available to the NGINX pod and then having NGINX serve it directly. (Right now, the request is sent by NGINX to UWSGI, who has access to the file off the container image, and then passes it back to NGINX. I believe when uWSGI passes the response, it types it as `text/html` since that's what NGINX always gets, even with `mime.types` explicitly included and with default type set to svg. This is necessary right now because NGINX pod doesn't have the galaxy codebase and static contents are not on the shared filesystem.)
[...]

PR #331, opened 2021-09-30
Make NGINX serve static content directly

almahmoud:
Fixes https://github.com/galaxyproject/galaxy-helm/issues/330 

`.Values.nginx.galaxyStaticDir` only affects where the static dir is mounted in the NGINX pod and where the `nginx.conf` expects to look. It is not related to the location in the Galaxy container image itself, which remains hardcoded.


almahmoud:
P.S.: Merge the next PR https://github.com/galaxyproject/galaxy-helm/pull/332 first to fix the linting CI

ksuderman (approved):

PR #354, opened 2022-03-07
Replace container mapper rules with TPV

nuwang:


afgane (approved):

afgane:
It replaces it. It sill allows for local rules (https://github.com/galaxyproject/galaxy-helm/pull/354/files#diff-86e6e8118f9d5ad6d181dd2e12c268062e9a66f5ef98bd0cc44b93661d08e9b2R479) and also relies on global rules (https://github.com/galaxyproject/galaxy-helm/pull/354/files#diff-86e6e8118f9d5ad6d181dd2e12c268062e9a66f5ef98bd0cc44b93661d08e9b2R476) so not every installation needs to make these decisions repeatedly.

PR #359, opened 2022-05-22
Added support for loading refdata from s3

nuwang:


afgane (approved):
I've deployed this code in a number of scenarios at this point and it's all largely worked. 

I've just not been able to deploy it on a local MacOS using Rancher Desktop/k3s with the `driver-registrar` container throws error saying that `kubelet-registration-path` is a required parameter. The [chart is supplying it](https://github.com/ctrox/csi-s3/pull/75/files#diff-d8d2b077c14e76fd1ba61f7d452a2ac56092bb9f4a67c2c3a1a92aec9a14028aR74) so seems it's an issue with the Mac env. 

Before merging, it may make sense to first merge #367, which updates K8s version and removes the CVMFS test so the tests here will pass?

ksuderman (approved):
LGTM

PR #367, opened 2022-07-29
Remove use of beta APIs used by older K8s versions

afgane:
With these changes, specifically the change to CronJob (https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/), the minimum supported Kubernetes version for this chart will be 1.21. Given that 1.21 reached end-of-life for maintenance support in June 2022 (https://endoflife.date/kubernetes), this seems reasonable.

Removed the test that included CVMFS because it's no longer compatible with the newer K8s version; will be replaced as part of #359 with S3fs instead.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,4 +1,4 @@
-name: Linting and deployment tests (w/ & w/out CVMFS) on K3S
+name: Linting and deployment test on K3S
 on:
   push:
     branches:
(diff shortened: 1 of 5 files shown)


afgane:
I've deployed Galaxy with this change all the way from K8s v1.21 through v1.24 and all the services came up as expected.

nuwang (approved):

PR #371, opened 2022-08-04
Remove the galaxy-cvmfs-csi dependency.

ksuderman:
The cvmfs-csi should not be required any longer since moving to S3FS.


nuwang (approved):

nuwang:
Looks like this change was already included in the previous PR, so closing.

PR #374, opened 2022-08-14
Add support for Galaxy 22.05

nuwang:
* Switches from uwsgi to gunicorn
* Adds celery and celery-beat
* Enables tusd by default
* Postgres operator updated
* Pins nginx container image

### DIFF
--- a/galaxy/templates/_helpers.tpl
+++ b/galaxy/templates/_helpers.tpl
@@ -163,6 +163,7 @@ until [ -f /galaxy/server/config/mutable/init_mounts_done_{{$.Release.Revision}}
 {{ if .Values.setupJob.downloadToolConfs.enabled }}
 until [ -f /galaxy/server/config/mutable/init_clone_done_{{$.Release.Revision}} ]; do echo "waiting for refdata copying"; sleep 1; done;
 {{- end }}
+echo "Initialization waits complete";
 {{- end -}}
 
 {{/*
@@ -221,6 +222,10 @@ Define pod env vars
                 secretKeyRef:
                   name: "{{ .Release.Name }}-galaxy-secrets"
                   key: "galaxy-config-id-secret"
+            - name: PYTHONPATH
+              value: /galaxy/server/lib
+            - name: GALAXY_CONFIG_FILE
+              value: /galaxy/server/config/galaxy.yml
 {{- end -}}
 
 
--- a/galaxy/templates/deployment-celery-beat.yaml
+++ b/galaxy/templates/deployment-celery-beat.yaml
@@ -146,12 +146,6 @@ spec:
             {{- end }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
-        # - name: nginx
-        #   image: nginx
-        #   lifecycle:
-        #     preStop:
-        #       exec:
-        #         command: ["/usr/sbin/nginx","-s","quit"]
       volumes:
         - name: galaxy-conf-files
           {{- if .Values.useSecretConfigs }}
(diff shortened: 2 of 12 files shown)


galaxy/values.yaml
```diff
@@ -6,8 +6,8 @@ nameOverride: ""
 fullnameOverride: ""
 
 image:
-  repository: galaxyproject/galaxy-min
-  tag: 21.09
+  repository: galaxy/galaxy-min
```
afgane, line 9: This will require a change in the galaxy auto-build Action, probably best to do as part of this PR https://github.com/galaxyproject/galaxy/pull/14449.

galaxy/templates/deployment-celery-beat.yaml
```diff
@@ -0,0 +1,207 @@
...
+            {{- tpl (.Values.extraVolumeMounts | toYaml | nindent 12) . }}
+            {{- end }}
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
+        # - name: nginx
+        #   image: nginx
+        #   lifecycle:
+        #     preStop:
+        #       exec:
+        #         command: ["/usr/sbin/nginx","-s","quit"]
```
afgane, line 154: Why have this here? If left, we should add a comment when it should be uncommented.

galaxy/templates/deployment-celery.yaml
```diff
@@ -0,0 +1,207 @@
...
+            {{- tpl (.Values.extraVolumeMounts | toYaml | nindent 12) . }}
+            {{- end }}
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
+        # - name: nginx
+        #   image: nginx
+        #   lifecycle:
+        #     preStop:
+        #       exec:
+        #         command: ["/usr/sbin/nginx","-s","quit"]
```
afgane, line 154: Same comment as in the `celery-beat` template.

galaxy/values.yaml
```diff
@@ -86,6 +90,57 @@ jobHandlers:
     failureThreshold: 30
     timeoutSeconds: 5
 
+celery:
+  concurrency: 2
+  logLevel: "DEBUG"
```
afgane, line 95: I recall Celery being chatty at `DEBUG`. Is this still the case? Should we bump this to `INFO`?

galaxy/templates/_helpers.tpl
```diff
@@ -163,6 +163,7 @@ until [ -f /galaxy/server/config/mutable/init_mounts_done_{{$.Release.Revision}}
 {{ if .Values.setupJob.downloadToolConfs.enabled }}
 until [ -f /galaxy/server/config/mutable/init_clone_done_{{$.Release.Revision}} ]; do echo "waiting for refdata copying"; sleep 1; done;
 {{- end }}
+echo "Initialization waits complete";
```
afgane, line 166: ♥

afgane (approved):
This looks great! The only thing that looks like is needed before it'll work is to move our use of Docker Hub to `galaxy` instead of `galaxyproject`.

galaxy/templates/deployment-celery-beat.yaml
nuwang: Totally missed this. Seems to have been copied over from the web handler. Removed all.

nuwang:
Changes done. I'll follow up with a separate PR which adds support for the rabbitmq operator.

galaxy/values.yaml
nuwang: Done

PR #378, opened 2022-08-20
Add support for rabbitmq operator

nuwang:


afgane (approved):
This is a nice and compact addition; comes up as expected. 

The only annoyance now is that the chart no longer works on M1 Mac's until an ARM image for RabbitMQ is available (ref https://github.com/bitnami/charts/issues/7305).

This supersedes #152.

PR #379, opened 2022-08-30
Tag releases

ksuderman:
Add an action to the `package.yml` workflow to create a tag and GitHub release when then chart is deployed to the Helm Chart repository.


afgane:
This is certainly better than overwriting already published releases with each PR merge, but being able to simply get latest packaged code is also convenient. So how about when no labels are added to a PR, we still automatically package the chart as `galaxy-dev`?

ksuderman:
We (I) will have to look at the `helm_ci` action since it bumps the version number **and** does the `helm package`.

ksuderman:
It is easier (for me at least) to simply re-roll this PR rather than struggling through a rebase.

PR #383, opened 2022-10-04
Use manage_db.sh init to create the database on first boot

ksuderman:
The `create_db.sh` script was removed from 22.05 and is replaced with `manage_db.sh init`.


ksuderman:
> [`job-init-db.yaml`](https://github.com/galaxyproject/galaxy-helm/blob/9231aefb626bada1cb760afa58fa83772cd07a79/galaxy/templates/job-init-db.yaml#L38) also needs to be updated.
> 

They are the same file. It was renamed in 4e33306cd04fcdbc1d7d311c09c52bb688275607

afgane (approved):
Ahh, sorry about that. That's what I get for relying on github search only...

PR #385, opened 2022-10-07
Upgrade postgres operator and use latest k3s version (1.26) for tests

afgane:


afgane:
Looks like using K8s 1.25 will have to wait for [postgres-operator 1.9](https://github.com/zalando/postgres-operator/milestone/12). While I'm not seeing the [same postgres startup issue as reported when running the 1.8.2 operator on K8s v1.25](https://github.com/zalando/postgres-operator/issues/1999), our db-init job keeps crash looping and reporting authentication failed for galaxydbuser in the log. Meanwhile, the chart continues to deploy fine with the operator on 1.24 as well as without the operator. Other than that, I don't have other leads as to why it won't run on 1.25.

afgane:
Postgres Operator v1.9 has been released (https://github.com/zalando/postgres-operator/releases/tag/v1.9.0), so hopefully our tests pass now.

Meanwhile, K8s 1.26 has been released.

PR #400, opened 2022-12-15
Completes docs on using external files

pcm32:
A small clarification that when using external files you still need to include some part for them on the extra file mappings of the values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,8 @@ extraFileMappings:
 1. Creating a symbolic link in the chart directory to the external file, or
 2. using `--set-file` to specify the contents of the file. E.g:
    `helm upgrade --install galaxy cloudve/galaxy -n galaxy --set-file extraFileMappings."/galaxy/server/static/welcome\.html".content=/home/user/data/welcome.html`
+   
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -261,6 +261,8 @@ extraFileMappings:
 1. Creating a symbolic link in the chart directory to the external file, or
 2. using `--set-file` to specify the contents of the file. E.g:
    `helm upgrade --install galaxy cloudve/galaxy -n galaxy --set-file extraFileMappings."/galaxy/server/static/welcome\.html".content=/home/user/data/welcome.html`
+   
+Loading the files from external locations still requires a minimal entry in the `extraFileMappings` for that file, as shown above but with no `content:` part. Otherwise, the setup doesn't know on which pods to apply this.
```
nuwang, line 265: Can we add those as extra set properties, to be more inline with this particular example?
```
extraFileMappings."/galaxy/server/static/welcome\.html".applyToWeb=true
```

pcm32: Ahhh, cool, how come I didn't thought of that. Sure, will do.

Issue #399, opened 2022-12-15
Run out of file locks after an hour of running (around ~100 jobs)


pcm32:
After around an hour of running I'm starting to get all sort of file locking errors in the main job handler:

```
galaxy.tool_util.deps.containers ERROR 2022-12-15 10:34:46,234 [pN:job_handler_0,p:8,tN:KubernetesRunner.work_thread-1] Could not get container description for tool 'toolshed.g2.bx.psu.edu/repos/ebi-gxa/scanpy_compute_graph/scanpy_compute_graph/1.8.1+2+galaxy0'
Traceback (most recent call last):
  File "/galaxy/server/lib/galaxy/tool_util/deps/containers.py", line 320, in find_best_container_description
    resolved_container_description = self.resolve(enabled_container_types, tool_info, **kwds)
  File "/galaxy/server/lib/galaxy/tool_util/deps/containers.py", line 351, in resolve
    container_description = container_resolver.resolve(
  File "/galaxy/server/lib/galaxy/tool_util/deps/container_resolvers/mulled.py", line 557, in resolve
    name = targets_to_mulled_name(
  File "/galaxy/server/lib/galaxy/tool_util/deps/container_resolvers/mulled.py", line 361, in targets_to_mulled_name
    tags = mulled_tags_for(namespace, target.package_name, resolution_cache=resolution_cache, session=session)
  File "/galaxy/server/lib/galaxy/tool_util/deps/mulled/util.py", line 127, in mulled_tags_for
    if not _namespace_has_repo_name(namespace, image, resolution_cache):
  File "/galaxy/server/lib/galaxy/tool_util/deps/mulled/util.py", line 104, in _namespace_has_repo_name
    cached_namespace = preferred_resolution_cache.get(cache_key)
  File "/galaxy/server/.venv/lib/python3.10/site-packages/beaker/cache.py", line 322, in get
    return self._get_value(key, **kw).get_value()
  File "/galaxy/server/.venv/lib/python3.10/site-packages/beaker/container.py", line 332, in get_value
    self.namespace.acquire_read_lock()
  File "/galaxy/server/.venv/lib/python3.10/site-packages/beaker/container.py", line 206, in acquire_read_lock
    self.access_lock.acquire_read_lock()
  File "/galaxy/server/.venv/lib/python3.10/site-packages/beaker/synchronization.py", line 154, in acquire_read_lock
    x = self.do_acquire_read_lock(wait)
[...]

PR #402, opened 2022-12-16
Sets the mulled resolution cache lock dir to local

pcm32:
The mulled cache lock dir is often accessed and it doesn't need to be on the shared file system. Then each galaxy process that needs it will have it's own and performance on local disk should be better than on the shared file system.

Helps with #399 .

### DIFF
--- a/galaxy/values.yaml
+++ b/galaxy/values.yaml
@@ -409,6 +409,7 @@ configs:
       interactivetools_base_path: "{{$host := index .Values.ingress.hosts 0}}{{$path := index $host.paths 0}}{{$path.path}}"
       id_secret:
       logo_src: 'static/favicon.png'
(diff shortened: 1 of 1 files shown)


nuwang (commented):
@pcm32 Does the /galaxy/server/local directory need to be created beforehand? Or will the entire path be created automatically?

pcm32:
On the current deployment, it created it on its own, but it was just a single level directory. I will restart my deployment with this path to confirm that the behaviour is recursive.

pcm32:
I can confirm that the directories requested get created by some process without the helm chart having to take care of it.

nuwang (approved):

nuwang:
Thanks @pcm32!

PR #405, opened 2022-12-24
Make necessary changes for TPV 2.0

nuwang:


afgane:
Will this work with with older Galaxy versions? I cannot find where TPV version is pinned so wondering how does an admin update from <2.0.

nuwang:
Yes. The runner parameter will be ignored in older versions. The TPV version is constrained in Galaxy (https://github.com/galaxyproject/galaxy/blob/c54ec2f18d7a8d5fa7a73710c550ad24ecda594b/lib/galaxy/dependencies/conditional-requirements.txt#L16), but unconstrained in the container (https://github.com/galaxyproject/galaxy-docker-k8s/blob/3d0ccac59107841ff28acc2eb60e433e967fc7eb/files/k8s-requirements.txt#L3).

afgane (approved):

PR #413, opened 2023-02-21
Set docker_enabled only once

afgane:
At the very least, having the `enable_docker` setting set twice in the default values file seems undesirable. On top of that, I don't think we run the chart with docker available much any more so disabling it by default seems like a reasonable option.

### DIFF
--- a/galaxy/values.yaml
+++ b/galaxy/values.yaml
@@ -374,7 +374,7 @@ configs:
       environments:
         tpv_dispatcher:
           container_monitor: false
-          docker_enabled: false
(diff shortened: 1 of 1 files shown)


pcm32:
I'm not sure I'm getting this. With this jobs need to have a different way of resolving tool dependencies that is not though containers? Or this is something else?

afgane:
You're right. I misunderstood what the `enable_docker` variable did so we do need it to set to `true` by default. But I still don't think we need it set twice in our values file.

nuwang (approved):

PR #414, opened 2023-03-06
Upgrade Galaxy version to 23.0

afgane:
The auto-built image doesn't work out of the box atm because the version of TPV is wrong (see https://github.com/galaxyproject/galaxy/pull/15647#issuecomment-1453926643), but using a manually built image confirms the upgrade is working.

There are two issues/discussion points I'd like to raise:
1. Should we consider versioning the chart in some way that mimics the Galaxy versions? If not, should this PR invoke a major version bump on the chart or something else?
2. The default TPV database seems too generous with the resources for many tools for smaller instances. This is particularly evident during the chart development lifecycle where it's difficult to get many jobs to run on small dev envs (eg, [FastQC requests 8 CPUs](https://github.com/galaxyproject/tpv-shared-database/blob/59db8c3d3af50a802005a0338a0771d9132df570/tools.yml#LL357-L358)). Perhaps we could add an ability in TPV to scale all the default resources by some factor that would then make the those resource requests more acceptable to a local environment?

### DIFF
--- a/galaxy/Chart.yaml
+++ b/galaxy/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: galaxy
 type: application
 version: 5.5.0
-appVersion: "22.05"
+appVersion: "23.0"
 description: Chart for Galaxy, an open, web-based platform for accessible, reproducible, and transparent computational biomedical research.
 icon: https://galaxyproject.org/images/galaxy-logos/galaxy_project_logo_square.png
 dependencies:
--- a/galaxy/values.yaml
+++ b/galaxy/values.yaml
@@ -6,7 +6,7 @@ fullnameOverride: ""
 
 image:
   repository: quay.io/galaxyproject/galaxy-min
-  tag: 22.05
+  tag: "23.0"  # Value must be quoted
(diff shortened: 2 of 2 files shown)


nuwang:
Documenting the discussion here for posterity.

1. The general helm chart convention is to use appVersion to track the underlying app's version, and to use semantic versioning for the helm chart, so that the two can be independently versioned. Therefore, the key basis for whether the chart needs a major version bump would be whether the chart can be
a) Smoothly upgraded without hiccups
b) Has no breaking changes to the chart interface
Therefore, this particular case would be a minor version bump, despite the major appVersion bump.

2. We can probably using clamping as a workaround for this: https://total-perspective-vortex.readthedocs.io/en/latest/topics/tpv_by_example.html#clamping-resources
In particular, development setups would have to set something like
```console
helm install .... --set jobs.rules.tpv_rules_local\.yml.destinations.k8s.max_cores=2
```

afgane:
With https://github.com/galaxyproject/galaxy/pull/15710 that fixes TPV versioning, this is ready for review/merge.

nuwang (approved):

PR #419, opened 2023-03-16
Add a post-install hook to restart cvmfs csi pods

afgane:
As a workaround for what's seemingly a bug in the CVMFS-CSI (see https://github.com/cvmfs-contrib/cvmfs-csi/issues/70), restarting the nodeplugin pods after the chart has installed seems to fix the volume mount issue so add a post-install hook that will kill the relevant nodeplugin pods. 

For the hook to work properly, must install the chart with `--wait` flag.

### DIFF
--- a/galaxy/templates/configmap-cvmfs-fix.yaml
+++ b/galaxy/templates/configmap-cvmfs-fix.yaml
@@ -7,10 +7,12 @@ metadata:
 data:
   cvmfs-fix.sh: |
     sleep 10;
-    while [ "$(kubectl get pods -n {{ .Release.Namespace }} -l 'app=cvmfscis' -l 'component=nodeplugin' -o custom-columns=STATUS:.status.phase --no-headers)" != "Running" ]; do
-      echo "Waiting on nodeplugin pod to enter 'Running' status.";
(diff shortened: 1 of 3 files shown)


galaxy/templates/configmap-cvmfs-fix.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-configmap-cvmfs-fix
+  labels:
+    {{- include "galaxy.labels" . | nindent 4 }}
+data:
+  cvmfs-fix.sh: |
+    sleep 10;
+    while [ "$(kubectl get pods -n {{ .Release.Namespace }} -l 'app=cvmfscis' -l 'component=nodeplugin' -o custom-columns=STATUS:.status.phase --no-headers)" != "Running" ]; do
```
nuwang, line 10: Is the app label correct? Shouldn't it be `cvmfscsi` instead?
Also, will this work if there are multiple nodes and get pods returns multiple lines for running pods?

afgane: Both good catches. Should be fixed now.

PR #421, opened 2023-03-25
RabbitMQ protocol and port configurable as variables

yegortokmakov:
A TLS-enabled RabbitMQ cluster is served on port 5671 by default. This helm should support port and protocol customization via helm values.

I've set the defaults to `amqp` and `5672` to keep the current behavior.


pcm32:
Thanks, I agree that those values should be configurable. Do you know if the underlying operator or chart that is deploying rabbitmq will pick up these values as well?

ksuderman:
Thanks for the PR @yegortokmakov 

@pcm32 No, these values simply inform Galaxy how the RabbitMQ cluster has been configured.

PR #423, opened 2023-04-10
Add expression tools

ksuderman:
Add the `param_value_from_file` tool to a new section `Expression Tools`.

**Q:** Do we want to add the other expression tools (`expression_macros` and `pick_value`) at the same time?

### DIFF
--- a/galaxy/files/configs/tool_conf.xml
+++ b/galaxy/files/configs/tool_conf.xml
@@ -38,6 +38,9 @@
     <tool file="${model_tools_path}/build_list.xml" />
     <tool file="${model_tools_path}/extract_dataset.xml" />
   </section>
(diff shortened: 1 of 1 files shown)


afgane:
Those are installed from the toolshed, which I've initiated in
https://github.com/galaxyproject/usegalaxy-tools/pull/595 so once we
resolve the migration mentioned there (I spoke with JohnC about doing it
together after I'm back next week), the tools will be included.

ksuderman:
> Those are installed from the toolshed, which I've initiated in [galaxyproject/usegalaxy-tools#595](https://github.com/galaxyproject/usegalaxy-tools/pull/595)
> […](#)
Ah, that's where it is.  I couldn't find it when searching the toolshed.

afgane (approved):

PR #424, opened 2023-04-10
Fix copying default tools

### LINKED ISSUE #418: On Helm upgrade, Galaxy default tools don't get updated from a new galaxy container
On subsequent runs of `helm upgrade...` of the chart, when the galaxy image changes, the default tools in `/galaxy/server/database/tools` don't get updated from the newer ones in the container.

Also, there is a repeated `/galaxy/server/database/tools/tools` directory with the same default tools again.

Should this be added in the init-mounts perhaps?

ksuderman:
The default tools from the container were being copied to the wrong location on `helm upgrade`.

Fixes #418

### DIFF
--- a/galaxy/templates/_helpers.tpl
+++ b/galaxy/templates/_helpers.tpl
@@ -161,7 +161,7 @@ cp -anL /galaxy/server/config/sanitize_allowlist.txt /galaxy/server/config/mutab
 cp -anL /galaxy/server/config/data_manager_conf.xml.sample /galaxy/server/config/mutable/shed_data_manager_conf.xml;
 cp -anL /galaxy/server/config/tool_data_table_conf.xml.sample /galaxy/server/config/mutable/shed_tool_data_table_conf.xml;
 cp -aruL /galaxy/server/tool-data {{.Values.persistence.mountPath}}/;
-cp -aruL /galaxy/server/tools {{.Values.persistence.mountPath}}/tools | true;
+cp -aruL /galaxy/server/tools {{.Values.persistence.mountPath}} | true;
 echo "Done" > /galaxy/server/config/mutable/init_mounts_done_{{.Release.Revision}};
 {{- end -}}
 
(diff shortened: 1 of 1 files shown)


galaxy/templates/_helpers.tpl
```diff
@@ -161,7 +161,7 @@ cp -anL /galaxy/server/config/sanitize_allowlist.txt /galaxy/server/config/mutab
 cp -anL /galaxy/server/config/data_manager_conf.xml.sample /galaxy/server/config/mutable/shed_data_manager_conf.xml;
 cp -anL /galaxy/server/config/tool_data_table_conf.xml.sample /galaxy/server/config/mutable/shed_tool_data_table_conf.xml;
 cp -aruL /galaxy/server/tool-data {{.Values.persistence.mountPath}}/;
-cp -aruL /galaxy/server/tools {{.Values.persistence.mountPath}}/tools | true;
+cp -aruL /galaxy/server/tools {{.Values.persistence.mountPath}}/ | true;
```
almahmoud, line 164: Should this also be `|| true` ?

```suggestion
cp -aruL /galaxy/server/tools {{.Values.persistence.mountPath}}/ || true;
```

ksuderman: Good catch. Do we want/need that on the other cp commands?

almahmoud: I guess it wouldn't hurt but honestly I don't even remember what scenario we were accommodating with this one .

galaxy/templates/_helpers.tpl
```diff
@@ -161,7 +161,7 @@ cp -anL /galaxy/server/config/sanitize_allowlist.txt /galaxy/server/config/mutab
 cp -anL /galaxy/server/config/data_manager_conf.xml.sample /galaxy/server/config/mutable/shed_data_manager_conf.xml;
 cp -anL /galaxy/server/config/tool_data_table_conf.xml.sample /galaxy/server/config/mutable/shed_tool_data_table_conf.xml;
 cp -aruL /galaxy/server/tool-data {{.Values.persistence.mountPath}}/;
-cp -aruL /galaxy/server/tools {{.Values.persistence.mountPath}}/tools | true;
+cp -aruL /galaxy/server/tools {{.Values.persistence.mountPath}}/ || true;
```
nuwang, line 164:
```suggestion
cp -aruL /galaxy/server/tools {{.Values.persistence.mountPath}}/;
```

nuwang (approved):

nuwang (approved):

PR #430, opened 2023-04-26
Update s3-csi version to 0.35.2

nuwang:
xref: https://github.com/CloudVE/helm-charts/pull/4

### DIFF
--- a/galaxy/Chart.yaml
+++ b/galaxy/Chart.yaml
@@ -15,7 +15,7 @@ dependencies:
       - deploy-postgres
   - name: csi-s3
     repository: https://raw.githubusercontent.com/cloudve/helm-charts/master/
-    version: 0.31.3
+    version: 0.35.2
     condition: s3csi.deploy
     alias: s3csi
     tags:


yegortokmakov:
FYI, for some reason this new version of the chart is not handling the empty strings in the values well, so I had to define `csi-s3-secret` manually outside of the `csi-s3` chart. The error I was getting: 
```
error validating "": error validating data: [unknown object type "nil" in Secret.stringData.accessKeyID, unknown object type "nil" in Secret.stringData.secretAccessKey]
```
and should be fixed by defining default values in the chart itself.

nuwang:
Yeah, I just remembered that I made two upstream fixes that were never merged:
https://github.com/yandex-cloud/k8s-csi-s3/pull/11
https://github.com/yandex-cloud/k8s-csi-s3/pull/20

Without the second one in particular, we would have to manually package the chart and force the s3fs mounter. Last time I tried, the geesefs mounter has issues, in particular with caching: https://github.com/yandex-cloud/k8s-csi-s3/issues/18

PR #440, opened 2023-08-18
Only execute the hook-cvmfs-fix if CVMFS is actually enabled.

### LINKED ISSUE #439: Crashes without Refdata on CVMFS
I currently try to deploy the helm chart without reference data.  When I do, https://github.com/galaxyproject/galaxy-helm/blob/master/galaxy/templates/hook-cvmfs-fix.yaml this hook starts up and continues to wait for some cvmfs pods (ironically to kill them). As there is no pod, it runs indefinitely.

```
refdata.enabled: false
refdata.type: s3sci
s3csi.deploy: false
cvmfs.deploy: false
cvmfs.enabled: false
```

*Expected behaviour*:
If no cvmfs is used, then the pod for fixing cvmfs does not run. _Alternatively_ after no pod is found for 1-2 minutes, the pod finishes up.

Thank you!

ksuderman:
Adds a guard around the `hook-cvmfs-fix` template so it is only executed when CVMFS is enabled.  Mutually exclusive with [#441](https://github.com/galaxyproject/galaxy-helm/pull/441)

Closes #439

### DIFF
--- a/galaxy/templates/hook-cvmfs-fix.yaml
+++ b/galaxy/templates/hook-cvmfs-fix.yaml
@@ -1,3 +1,5 @@
+{{- if and .Values.refdata.enabled (eq .Values.refdata.type "cvmfs") }}
+  # Include the code you want to run when both conditions are met
 apiVersion: batch/v1
 kind: Job
 metadata:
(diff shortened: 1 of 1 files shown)


nuwang (approved):
LGTM!

PR #441, opened 2023-08-18
Remove CVMFS fix

### LINKED ISSUE #439: Crashes without Refdata on CVMFS

ksuderman:
Removes the  template.  Ensure it is no longer needed before merging.  Mutually exlusive with #440.  Closes #439


ksuderman:
Closing this since #440 was merged.

nuwang:
I didn't realize this was made in parallel, but we will need to verify that things work without the fix, and then close or not?

ksuderman:
Yeah, it was meant to be one or the other, but we can leave this open until we determine if the fix is still needed.

nuwang:
No longer needed because it was fixed upstream.

PR #445, opened 2023-10-11
Update app version to 23.1

ksuderman:
Update the Chart to use the `galaxy-min:23.1` Docker image.


nuwang (commented):
Maybe we can use appVersion for the tag from now on?

Also, the build has failed.

nuwang (approved):

afgane (approved):

PR #448, opened 2024-02-09
Cron job maintenance

### LINKED ISSUE #408: Maintenance job should also call maintenance.sh script and others
Currently the maintenance job https://github.com/galaxyproject/galaxy-helm/blob/master/galaxy/templates/cronjob-maintenance.yaml deletes the galaxy tmp. We should probably add there:

- execution of maintenance.sh script with a configurable days parameter.
- execution of the pgcleanup script to clean up older elements in the database.
- `/usr/bin/tmpwatch -v --mtime --dirmtime 7d` on the jobs working directory (or at least on `_cleared/`, which is the result of the resubmissions - the initial jobs that didn't work).

ksuderman:
Adds cron jobs that clean up the `jobs_directory`, datasets, and histories. The jobs are defined in the `values.yaml` file and templates generated from the `cronjob-maintenance.yaml` template.  The cron jobs are based on the [Ansible Playbook used on main](https://github.com/galaxyproject/usegalaxy-playbook/blob/main/env/main/host_vars/galaxy-web-05.tacc.utexas.edu.yml)

Since the `/galaxy/server/database/tmp` directory is cleaned up I am not sure the job that cleans the `tus_upload_store` is needed since it also lives in `/galaxy/server/database/tmp`, but I've included it for completeness.

The `pgcleanup.py` script writes logs to the `tmp` directory, but we should likely create a separate logs directory for these and then also prune the logs as needed.

Should we just call the `maintenance.sh` script, or a modified version of the script?

**NOTE** I have set the cron jobs to `runAsUser: 0` but we should probably run these as the Galaxy user once we have determined why `mutliqc` is creating files and directories as `nobody`.

This is untested as of yet.

Closes #408


nuwang (commented):
I love the idea of providing overridable cronjob templates with unlimited cron jobs. Given however, that the default cron jobs are all run daily, what if we just have a single bash script that is invoked by a single cronjob, which will in turn invoke the 3 cleanup commands sequentially? This can be overridden or replaced as desired.

galaxy/values.yaml
```diff
@@ -301,7 +361,8 @@ resources:
     memory: 7G
     ephemeral-storage: 10Gi
 
-nodeSelector: {}
+#nodeSelector: {}
+nodeSelector: "cloud:google:com/gke-nodepool: default-pool"
```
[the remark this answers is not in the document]
ksuderman: Don't know how that got in there...

ksuderman:
> what if we just have a single bash script that is invoked by a single cronjob, which will in turn invoke the 3 cleanup commands sequentially? This can be overridden or replaced as desired.

I was going to suggest the same thing, but I wanted to ensure the syntax for defining multiple cron jobs was working correctly first.  There is already a [maintenance.sh](https://github.com/galaxyproject/galaxy/blob/be90acce3afff88c3c991a638251d4cf820cc0e2/scripts/maintenance.sh#L1) script we could use as-is, or modify for our needs.

nuwang:
Makes sense to switch to the maintenance script if we can.

galaxy/values.yaml
```diff
@@ -267,6 +267,31 @@ extraEnv: []
...
+    securityContext:
+      runAsUser: 0
+    command:
+      - /usr/bin/chown
+    args:
+      - "-R"
+      - "galaxy"
+      - "/galaxy/server/database"
+#- Maintenance tasks that should be run periodically to prevent storage from being exhausted
+maintenance:
```
nuwang, line 283: Cam we make the maintenance jobs regular cronjobs as well, just with a "known" name? Something like:
```
cronJobs:
  maintenance:
    schedule: "0 * * * *"
    extraSettings:
       days: 7
    securityContext:
      runAsUser: 0
      command:
        - "/bin/bash"
        - "-c"      
      args:
        - "/galaxy/server/scripts/maintenance.sh"
        - "--no-dry-run"
        - "--days"
        - {{ .Values.cronJobs.maintenance.extraSettings.days | quote }}
```

And just tpl the `args` value in the cronjob template?

ksuderman: I guess we could.  Should we do the same for the tmp clean up script as well then?  I assume we may as well.

ksuderman:
Please ensure #453 has been merged before merging this.

nuwang (approved):
Looks good to me! Can this be taken out of draft?

galaxy/values.yaml
```diff
@@ -267,6 +267,53 @@ extraEnv: []
...
+    args:
+      - "/galaxy/server/scripts/maintenance.sh"
+      - "--no-dry-run"
+      - "--days"
+      - "{{ .Values.cronJobs.maintenance.extraSettings.days }}"
+  #- Remove files from the tmp directory that are older than the allowable wall time for a job
+  tmp:
+    schedule: "15 2 * * *"
+    extraSettings:
+      seconds: 604800
```
nuwang, line 300: The original code had a minus in front?
```suggestion
      seconds: -{{ (index .Values "configs" "job_conf.yml" "runners" "k8s" "k8s_walltime_limit" | default 604800) }}
```

galaxy/values.yaml
```diff
@@ -267,6 +267,53 @@ extraEnv: []
...
+    securityContext:
+      runAsUser: 0
+    command:
+      - /usr/bin/find
+    args:
+      - "{{ .Values.persistence.mountPath }}/tmp"
+      - "!"
+      - "-newermt"
+      - "{{ .Values.cronJobs.tmp.extraSettings.seconds }}"
```
nuwang, line 309:
```suggestion
      - "tpl {{ .Values.cronJobs.tmp.extraSettings.seconds }} $"
```

galaxy/values.yaml
ksuderman: The leading minus sign (less than) is not required here as that is what `-newermt` does anyway.  At least that is my understanding of the `find` documentation.  ChatGPT agrees (FWIW).

galaxy/values.yaml
```diff
@@ -267,6 +267,53 @@ extraEnv: []
...
+  chown:
+    schedule: "0 * * * *"
+    securityContext:
+      runAsUser: 0
+    command:
+      - /usr/bin/chown
+    args:
+      - "-R"
+      - "galaxy"
+      - "/galaxy/server/database"
```
nuwang, line 281: Why is the chown cronjob there? It doesn't seem right to me that it should be necessary.

ksuderman: Yeah, sorry.  I am in the process of reverting that right now....

galaxy/values.yaml
```diff
@@ -267,6 +267,53 @@ extraEnv: []
   # - name: EXAMPLE_ENV
   #   value: MY_VALUE
 
+#- CronJobs to perform periodic maintenacne tasks
```
nuwang, line 270:
```suggestion
#- CronJobs to perform periodic maintenance tasks
```

README.md
```diff
@@ -417,6 +417,10 @@ The Galaxy application can be horizontally scaled for the web, job, or workflow
 by setting the desired values of the `webHandlers.replicaCount`,
 `jobHandlers.replicaCount`, and `workflowHandlers.replicaCount` configuration options.
 
+## Cron jobs
+
+Two Cron jobs are defined to clean up Galaxy's database and the `tmp` directory.  By default these
+jobs run at 02:05 (the database maintenance script) and 02:15 (`tmp` directyory cleanup). Users
```
nuwang, line 423:
```suggestion
jobs run at 02:05 (the database maintenance script) and 02:15 (`tmp` directory cleanup). 
```

galaxy/values.yaml
```diff
@@ -267,6 +267,53 @@ extraEnv: []
...
+    args:
+      - "/galaxy/server/scripts/maintenance.sh"
+      - "--no-dry-run"
+      - "--days"
+      - "{{ tpl .Values.cronJobs.maintenance.extraSettings.days $ }}"
+    extraFileMappings:
+      /galaxy/server/scripts/maintenance.sh:
+        tpl: true
+        content: |
+          {{ .Files.Get "files/source/maintenance.sh" }}
```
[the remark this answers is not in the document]
ksuderman: Sorry, this push was only to test issues I am experiencing with GitHub at the moment and I guess I should have checked more closely what was being pushed.  The intent here was to overwrite the maintenance.sh script in the container with one that does not write to a log file.  However, after your comments on the logrotate PR I am in the process of creating a separate Docker image that can be used to run the maintenance tasks.

galaxy/values.yaml
```diff
@@ -443,7 +480,8 @@ configs:
       interactivetools_base_path: "{{$host := index .Values.ingress.hosts 0}}{{$path := index $host.paths 0}}{{$path.path}}"
       id_secret:
       mulled_resolution_cache_lock_dir: "/galaxy/server/local/mulled_cache_lock"
-      database_connection: postgresql://unused:because@overridden_by_envvar
+      database_connection: |-
+        {{ include "galaxy-postgresql.connection-string" .}}
```
nuwang, line 484: This was set to a constant on purpose as I recall, so that the credentials do not get embedded in a plaintext file.

README.md
```diff
@@ -417,21 +419,25 @@ The Galaxy application can be horizontally scaled for the web, job, or workflow
...
-`21.05` to `21.09` should be a matter of `helm upgrade my-galaxy cloudve/galaxy
---reuse-values --set image.tag=21.09`.
+```yaml
+cronJobs:
+  maintenance:
+    schedule: "30 6 * * *" # Execute the cron job at 6:30 UTC
+```
 
+To disable a cron job after Galaxy has been deployed simply set the schedule to a date that
+can never occur such as midnight on Februrary 30th:
```
nuwang, line 435: Better to add an enabled flag as we have for everything else?

galaxy/values.yaml
```diff
@@ -267,10 +267,47 @@ extraEnv: []
...
+      - "/galaxy/server/scripts/maintenance.sh"
+    args:
+      - "--no-dry-run"
+      - "--days"
+      - "{{ tpl .Values.cronJobs.maintenance.extraSettings.days $ }}"
+  #- Remove files from the tmp directory that are older than the allowable wall time for a job
+  tmpdir:
+    schedule: "15 2 * * *"
+    extraSettings:
+      last_modified: '{{ index .Values "configs" "job_conf.yml" "runners" "k8s" "k8s_walltime_limit" | default 604800 }}'
```
nuwang, line 292:
```suggestion
      lastModified: '{{ index .Values "configs" "job_conf.yml" "runners" "k8s" "k8s_walltime_limit" | default 604800 }}'
```

galaxy/values.yaml
```diff
@@ -267,10 +267,47 @@ extraEnv: []
...
+    securityContext:
+      runAsUser: 0
+    command:
+      - /usr/bin/find
+    args:
+      - "{{ .Values.persistence.mountPath }}/tmp"
+      - "!"
+      - "-newermt"
+      - "{{ tpl .Values.cronJobs.tmpdir.extraSettings.last_modified $ }} seconds ago"
```
nuwang, line 301:
```suggestion
      - "{{ tpl .Values.cronJobs.tmpdir.extraSettings.lastModified $ }} seconds ago"
```

galaxy/templates/cronjob-maintenance.yaml
```diff
@@ -1,47 +1,102 @@
...
+              name: {{ printf "%s-%s" (include "galaxy.fullname" $) (include "galaxy.getExtraFilesUniqueName" $key) }}
+            {{- end }}
+          {{- end }}
           restartPolicy: OnFailure
+{{- if $cronjob.extraFileMappings }}
+{{- range $name, $entry := $cronjob.extraFileMappings }}
+---
+apiVersion: v1
+metadata:
+  # Extract the filename portion only
```
nuwang, line 81: An example of extrafile mappings are not shown in values.yaml. Maybe add a comment?

README.md
ksuderman: Reading more about cron jobs,  maybe I should be using `.spec.suspend` here.  I'll play around with that and see if I get the desired behavior.

nuwang: Good idea to investigate suspend. Maybe we should add both.

galaxy/templates/cronjob-maintenance.yaml
ksuderman: Done.  I've added and example cron job and documented in the README.

galaxy/values.yaml
ksuderman: Per the discussion on Slack I will *resolve* this and leave as is until the Galaxy maintenance script properly support `GALAXY_CONFIG_OVERRIDE_*` variables.

README.md
ksuderman: Resolved in further README updates.

galaxy/values.yaml
ksuderman: Done.

README.md
nuwang: I think that adding an enabled flag would still be useful. Users who want to run the cronjob manually can still use the trick you suggested. Just that it feels weird to have to think of an invalid date for the simplest case of not wanting the cron job.

galaxy/templates/cronjob-maintenance.yaml
```diff
@@ -1,47 +1,112 @@
...
+  {{- include "galaxy.getFilenameFromPath" $name | nindent 2 }}: |
+  {{- if $entry.tpl }}
+    {{- tpl (tpl $entry.content $) $ | nindent 4 }}
+  {{- else }}
+    {{- $entry.content | nindent 4 }}
+  {{- end }}
+{{- end }}
+{{- end }}
+
+{{- end }}
```
nuwang, line 112: Needs a newline

nuwang (commented):
@ksuderman Looks like we have a conflict, and I've made some more minor comments, but other than that, this looks like it's good to go?

galaxy/templates/cronjob-maintenance.yaml
ksuderman: Added.

nuwang (approved):

PR #453, opened 2024-02-19
Ensure that one of the dependent jobs succeeded.

### LINKED ISSUE #452: Releases are not being tagged
The `tag-and-release` job in the workflow is not firing because the two job dependencies are mutually exclusive.

ksuderman:
The `tag-and-release` job depends on the `package-from-pr` or `package-from-manual` jobs succeeding.  However, the packaging jobs are mutually exclusive so we need to test if one, but not both, have succeeded.

See [Stack Overflow](https://stackoverflow.com/questions/66343052/github-actions-or-operator-for-needs-clause)

Closes #452

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -51,6 +51,10 @@ jobs:
     needs: [ package-from-pr, package-from-manual ]
     name: Create a tag and GitHub release for this version.
     runs-on: ubuntu-latest
+    if: |
+      always() 
+      && contains(needs.*.result, 'success')
+      && !contains(needs.*.result, 'failure')
     steps:
       - name: Tag and release
         run: |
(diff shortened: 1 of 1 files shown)


nuwang (approved):
Great! Thanks for this fix. Can we rerun this manually and rebuild the release history? Also, why is it missing the changelog despite --release-notes being present?

ksuderman:
No, this will only tag the latest commit/version.  I have a separate script I used to tag/release previous versions.  That script didn't include `--generate-notes` since the commit messages have been vague and consist almost entirely of "Automatic Version Bumping from X to X+1". We can regenerate them if needed though.

I'll add my `tag-and-release.sh` script to the repo in a separate PR.

nuwang:
So `--generate-notes` doesn't do the same thing as here?
image

Is there someway we can make it add the changelog? That would be perfect.

ksuderman:
No, it does not seem to when generated releases from past tags.  See the 5.8.0 and 5.9.0 releases I just generated.  **Note** that is just when generating release notes from previous tags/commits.  It should be find going forward.

nuwang:
Should be good to go then?

ksuderman:
I _think_ so...

PR #455, opened 2024-02-21
Add tag and release script

ksuderman:
This is the script I have been using to tag and release prior versions.


scripts/tag-and-release.sh
```diff
@@ -0,0 +1,26 @@
...
+	tag=v$(echo $line | awk '{print $2}')
+	# Get the actual date the commit was made
+	export GIT_COMMITTER_DATE=$(git show --format=%aD $commit | head -1)
+	# Tag the commit
+	echo "Tagging $commit as $tag $GIT_COMMITTER_DATE"
+	git checkout $commit
+	git tag -a -m "Automatic tagging of $tag" $tag
+	git push origin $tag
+	# Generate the release.
+	gh release create $tag --generate-notes --latest
```
nuwang, line 23: In the GUI release page, it's possible to select the previous release tag, so that release notes are generated in-between that and the latest. Is there a flag to do something like that in the commandline?

ksuderman: Ah, yes there is.  I missed it until now since its mixed in with the option to read the release notes from a file.

nuwang (approved):
Fabulous! Release notes look great now. Ok to merge?

ksuderman:
Yes, it should be good to go now.

PR #468, opened 2024-04-19
Add enable flag to cron jobs

ksuderman:
Adds an `enable` flag to each `cronJob` so they can be enabled/disabled.  Disabling a cron job is done by setting the schedule to `"0 0 30 2 * *"`, that is Feb 30th.  Since that date will never occur the job is effectively disabled.  Setting the `enable` flag to true resets the `schedule` to the value defined in the `values.yaml` file.

### DIFF
--- a/galaxy/templates/cronjob-maintenance.yaml
+++ b/galaxy/templates/cronjob-maintenance.yaml
@@ -10,6 +10,8 @@ spec:
   {{- if $cronjob.enabled }}
   schedule: {{ $cronjob.schedule | quote }}
   {{- else }}
+  # Set an impossible date so that the cronjob is still defined, but effectively disabled. 
+  #  This will allow the cronjob to be run manually if needed.
   schedule: "0 0 30 2 *"
   {{- end }}
   jobTemplate:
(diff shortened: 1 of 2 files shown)


galaxy/templates/cronjob-maintenance.yaml
```diff
@@ -7,7 +7,11 @@ metadata:
   labels:
     {{- include "galaxy.labels" $ | nindent 4 }}
 spec:
+  {{- if $cronjob.enabled }}
   schedule: {{ $cronjob.schedule | quote }}
+  {{- else }}
```
nuwang, line 12:
```suggestion
  {{- else }}
  # Set an impossible date so that the cronjob is still defined, but effectively disabled. 
  #  This will allow the cronjob to be run manually if needed.
```

nuwang (commented):
Even better!

galaxy/templates/cronjob-maintenance.yaml
nuwang: An example of how to run it manually using kubectl or similar would also be helpful I think

ksuderman: There is an example in the README. Do we need one here as well?

nuwang: I had missed that - all good then.

nuwang (approved):

PR #314, opened 2021-08-13
Changes for not mounting entire NFS to each job

almahmoud:
Friends with https://github.com/galaxyproject/galaxy/pull/12345

PR #479, opened 2024-06-03
Roll back the NFS mount changes from PR #314

### LINKED ISSUE #476: File permission problems with restricted NFS mounts
After merging PR #314 jobs are failing due to file permission problems or file not found errors for tools installed from the tool shed.

ksuderman:
Reverts the NFS PVC mounts to the state they were prior to #314 being merged (Helm chart v 5.9.x)

Closes #476


nuwang (commented):
I don't think we should revert everything. We should just set tool-data and maybe tools to rw?

PR #481, opened 2024-06-29
Don't give read permission to the entire objects directory

### LINKED ISSUE #478: Cannot delete files/folders in the galaxy database owned by root users
Hello,
I have a similar problem that was mentioned in a previous post #398. Currently, I am inside the galaxy-job pod and then went to the database directory which looks like this:
```
galaxy/server/database$ ls
cache		      config	  deps		  objects	      shed_tools  tool-data  tool_search_index
celery-beat-schedule  cvmfsclone  jobs_directory  object_store_cache  tmp	  tools
```
I am trying to build Kraken database and do some other analysis after launching Galaxy locally and all the output files are being deposited to the `objects` folder. But when I look at the permission of all these folders, they are owned by root users:
```
/galaxy/server/database$ ls -la objects/
total 42
drwxr-xr-x 15 galaxy root 15 May 29 12:09 .
drwxrwxrwx 14 root   root 15 May 24 13:28 ..
drwxr-xr-x  3 galaxy root  3 May 29 12:09 0
drwxrwxrwt  5 galaxy root  5 May 29 12:09 1
drwxr-xr-x  4 galaxy root  4 May 29 12:09 2
drwxrwxrwt  6 galaxy root  6 May 29 12:09 4
drwxr-xr-x  3 galaxy root  3 May 29 12:09 5
drwxr-xr-x  3 galaxy root  3 May 29 12:09 6
drwxrwxrwt  3 galaxy root  3 May 29 12:00 7
drwxrwxrwt  6 galaxy root  6 May 29 12:09 8
drwxr-xr-x  4 galaxy root  4 May 29 12:09 b
drwxrwxrwt  4 galaxy root  4 May 29 12:09 c
drwxrwxrwt  3 galaxy root  3 May 29 11:59 e
drwxrwxrwt  4 galaxy root  4 May 29 12:09 f
drwxr-xr-x  3 galaxy root  3 May 29 12:08 _metadata_files
```

In some cases, I could delete some of these folders, but I could not delete some folders with files that were created by someone else and hold the root permission. So, attempting to delete those folders results in permission-denied errors.

Can you please suggest how to delete those files/folders with root permission outside the local galaxy users like me? or is there a way to change the permissions of these folders?

### LINKED ISSUE #480: Remove objects read permission
Galaxy 24.1.1 will include the fix that will allow us to remove read permission for the entire objects directory.

ksuderman:
Versions of Galaxy prior to 24.1.1 (to be released) contain a bug that did not mount the `extra_files` directories to jobs causing some tools to fail.  Once 24.1.1 is available we can merge this to restrict jobs so they can only read their inputs and only write to their output locations.

See [PR #18462](https://github.com/galaxyproject/galaxy/pull/18462)

Closes #480 
Closes #478

### DIFF
--- a/galaxy/values.yaml
+++ b/galaxy/values.yaml
@@ -10,7 +10,7 @@ image:
   #- Repository containing the Galaxy image.
   repository: quay.io/galaxyproject/galaxy-min
   #- Galaxy Docker image tag (generally corresponds to the desired Galaxy version)
-  tag: "24.0.2"  # Value must be quoted
+  tag: "24.1.1"  # Galaxy versions prior to 24.1.1 contain a bug mapping the extra_files directory
   #-  Galaxy image [pull policy](https://kubernetes.io/docs/concepts/configuration/overview/#container-images)
   pullPolicy: IfNotPresent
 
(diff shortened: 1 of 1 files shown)


ksuderman:
My simple test of running the variant calling workflow passes so I think this is safe to merge now.

nuwang (approved):
Great news! Thanks for testing.

PR #483, opened 2024-06-30
Update readme

ksuderman:
Add instructions for installing and configuring dnsmasq on Debian and Fedora based systems.


ksuderman:
This still needs testing on the above named systems before merging.

nuwang (approved):

PR #484, opened 2024-07-05
[TEST] downgrade k8s version to 1.24 

anilthanki:
This is mostly a test to see if we can run 5.5.0 with k8s version 1.24, to inform our update strategy. We don't intend to merge this, just piggy bagging on the CI.


nuwang:
Assuming this is no longer needed.

PR #485, opened 2024-07-10
Check that the appVersion matches version returned by /api/version

ksuderman:
I frequently forget to update the `appVersion` in the `Chart.yaml` when the Galaxy version changes.  This PR adds a test that compares the `appVersion` in the `Chart.yaml` with the version info returned by the `/api/version` endpoint and fails if they are not the same.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -64,3 +64,21 @@ jobs:
       - name: Print workflow handler log
         run: bash -c "kubectl logs -n galaxy $(kubectl -n galaxy get pods | grep -o '[^ ]*galaxy-workflow[^ ]*')"
         if: always()
+      - name: Print service info
+        run: kubectl get svc -n galaxy
+        if: always()
+      - name: Print nginx service info
+        run: kubectl describe svc -n galaxy galaxy-nginx
+        if: always()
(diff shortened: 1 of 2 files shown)


nuwang (approved):
This is great! Thanks for this fix.

.github/workflows/test.yaml
```diff
@@ -64,3 +64,21 @@ jobs:
...
         run: bash -c "kubectl logs -n galaxy $(kubectl -n galaxy get pods | grep -o '[^ ]*galaxy-workflow[^ ]*')"
         if: always()
+      - name: Print service info
+        run: kubectl get svc -n galaxy
+        if: always()
+      - name: Print nginx service info
+        run: kubectl describe svc -n galaxy galaxy-nginx
+        if: always()
+      - name: Check appVersion
+        if: always()
```
nuwang, line 74: What does always() do btw? Do we need it here?

ksuderman: The test will always run even if previous tests have failed.  It is likely not necessary here, and actually the two tests that do nothing by print the output of `kubectl` could be folded into the `appVersion` check.  We likely want the `appVersion` check to always run to catch as many problems as possible at once.

PR #486, opened 2024-07-11
Fix s3csi path & readme

bravo325806:
I found no subpath when using s3csi. The readme refdata.type is for using s3csi not s3fs


ksuderman:
Thanks for the PR.  We are investigating as the `subPath` should be found.

bravo325806:
Hi ksuderman

This error occurs if there are subpath, and I use no subpath only see the byhand managed two folder
``` Error: failed to create subPath directory for volumeMount "refdata-gxy" of container "galaxy-init-mounts" ```

I found older versions don't have subpath. If the path of s3 has not changed from the old version to the current version, there should be no need to add a new subpath.

nuwang:
@bravo325806 Can you clarify what you meant by "I found older versions don't have subpath"? Are you trying to upgrade or is this a fresh install?

The subPath is not cvmfs specific and is needed for the bucket too: http://biorefdata.s3-website-ap-southeast-2.amazonaws.com/#galaxy/v1/

bravo325806:
@nuwang Sorry I didn't make this part clear. I used v5.1.0
https://github.com/galaxyproject/galaxy-helm/blob/v5.1.0/galaxy/templates/deployment-web.yaml#L138-L139
Now I have upgrade to v5.15.0, but what I see in the new version is that there is a subpath option https://github.com/galaxyproject/galaxy-helm/blob/v5.15.0/galaxy/templates/deployment-web.yaml#L135-L137

Because values.yaml
The prefix /galaxy/v1/data.galaxyproject.org is already preset
https://github.com/galaxyproject/galaxy-helm/blob/v5.15.0/galaxy/values.yaml#L453

Ｍaybe we can only change values.yaml?

ksuderman:
@bravo325806 I made a PR to your PR as jobs that are run also mount the refdata PV and should not include the `subpath`  when using the s3csi.

Also, that table of values in the README.md is generated by a script from comments in the `values.yaml` file, so I updated the comment to match your update to the readme so they don't get lost the next time we generate that table.

As soon as you merge those changes we can go ahead and merge your PR.

bravo325806:
@ksuderman I have merged your PR, Thanks

nuwang (commented):
Thanks both for persevering with these fixes + testing. The changes lgtm!

Issue #494, opened 2024-08-06
`post-install-cvmfs-fix` dosent-work when CVMFS is deployed as external dependency 


NilsKrattinger:
Hi, 

It seems that the `post-install-cvmfs-fix` is not able to terminate / run when `cvmfs` is deployed in another namespace.
I think this is because the namespace is retrieved from the `.Release`. in `configmap-cvmfs-fix` [example](https://github.com/galaxyproject/galaxy-helm/blob/91e11e0dfbaae3f48e04636eeaf42a7ca40cd890/galaxy/templates/configmap-cvmfs-fix.yaml#L11).

I don't have the history of the reason for this workaround, and I don't know if it's necessary even when CVMFS is deployed cluster-wide (I think it is).

If the fix is necessary, could we set the CVMFS namespace as a value?  This would probably require a dedicated SA with permissions on the CVMFS namespace.

If not, should the fix be conditional on `cvmfs.deploy` ?

If you could point me in the right direction here, I'd be happy to do this PR!

Pod Logs : 
```
 kubectl logs -n galaxy-dev dev01-post-install-cvmfs-fix-job-mqr62
---
Waiting on nodeplugin pod to enter 'Running' status. Currently ''.
Waiting on nodeplugin pod to enter 'Running' status. Currently ''.
Waiting on nodeplugin pod to enter 'Running' status. Currently ''.
Waiting on nodeplugin pod to enter 'Running' status. Currently ''.
Waiting on nodeplugin pod to enter 'Running' status. Currently ''.
Waiting on nodeplugin pod to enter 'Running' status. Currently ''.
Waiting on nodeplugin pod to enter 'Running' status. Currently ''.
Waiting on nodeplugin pod to enter 'Running' status. Currently ''.
Waiting on nodeplugin pod to enter 'Running' status. Currently ''.
Waiting on nodeplugin pod to enter 'Running' status. Currently ''.
Waiting on nodeplugin pod to enter 'Running' status. Currently ''.
Waiting on nodeplugin pod to enter 'Running' status. Currently ''.
Waiting on nodeplugin pod to enter 'Running' status. Currently ''.
```

ksuderman:
Yeah, it looks like the fix assumes CVMFS will be in the `.Release.Namespace`. We could do with better logic when it comes to locating the CVMFS `nodeplugin`.  It might also be a good idea to make the fix conditional on a flag set in the `values.yaml` file; something like `cvmfs.enable-fix=false`.

[...]

PR #498, opened 2024-08-12
Make CVMFS fix configurable

NilsKrattinger:
Add a option to not deploy the CVMFS-fix. 
~Add ability to restart nodeplugin even if deployed on other namespace~ 
Add condition to not create the cvmfs-fix configmap if CVMFS isn't used / deployed as CSI for refdata. 

Close [#494](https://github.com/galaxyproject/galaxy-helm/issues/494).
~Impact #441.~


galaxy/templates/rbac-fix-cvmfs.yaml
```diff
@@ -0,0 +1,42 @@
...
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "list","watch","delete"]
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ include "galaxy.fullname" . }}-cvmfs-fix
+  namespace: {{ .Values.cvmfs.externalCvmfsNamespace }}
```
nuwang, line 31: In general, I think it's preferable to constrain the helm chart to a single namespace. We don't want to expand the security footprint where it reaches across to other namespaces. Therefore, I think that if someone wants to create this service account, they do so manually, and we just document it.

nuwang (commented):
Thanks for this fix @NilsKrattinger. My only suggestion is that we add the `deployPostInstallFix` flag only, and remove the `externalCvmfsNamespace` option. This fix was never meant to be permanent, just a temporary workaround till the upstream csi chart is fixed. I think it's preferable to simply document the steps or manually restart the csi pods instead.

NilsKrattinger:
@nuwang, I have removed the ability to restart the external Nodepluing CVMFS. 

Changes:
- Add a option to not deploy the CVMFS-fix. ``
- Add condition to not create the cvmfs-fix configmap if CVMFS isn't used / deployed as CSI for refdata.

Close #494.

nuwang:
@NilsKrattinger Thanks for the changes. @ksuderman is also investigating whether this may have been fixed upstream, and hopefully he has some good news to share.

PR #501, opened 2024-09-11
Update dependency charts; save outputs to working dir

afgane:
I tested the updated deps on a K32 (1.30) cluster and worked as expected.

The outputs_to_working_directory option was a conclusion from a recent discussion.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "ansible.python.interpreterPath": "/Users/ea/opt/anaconda3/bin/python"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "ansible.python.interpreterPath": "/Users/ea/opt/anaconda3/bin/python"
-}
--- a/galaxy/Chart.yaml
+++ b/galaxy/Chart.yaml
@@ -22,14 +22,14 @@ dependencies:
       - deploy-s3csi
   - name: galaxy-cvmfs-csi
     repository: https://raw.githubusercontent.com/cloudve/helm-charts/master/
-    version: 2.4.0
+    version: 2.5.0
     condition: cvmfs.deploy
     alias: cvmfs
     tags:
       - deploy-cvmfs
   - name: rabbitmq-cluster-operator
     repository: https://charts.bitnami.com/bitnami
-    version: 2.6.12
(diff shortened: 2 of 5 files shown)


.vscode/settings.json
ksuderman, line 1: This file should be in .gitignore?

ksuderman (changes requested):

.vscode/settings.json
afgane: Yes. That file should not have been committed...

afgane:
I don't understand what's going on with the tests so I hope it is transient. What worked a few hours ago on a VM now doesn't  so I'll just wait a bit for now.

```
Downloading postgres-operator from repo https://opensource.zalando.com/postgres-operator/charts/postgres-operator/
Save error occurred:  could not download postgres-operator-1.9.0.tgz: non-absolute URLs should be in form of repo_name/path_to_chart, got: postgres-operator-1.9.0.tgz
Error: could not download postgres-operator-1.9.0.tgz: non-absolute URLs should be in form of repo_name/path_to_chart, got: postgres-operator-1.9.0.tgz
```

galaxy/Chart.yaml
```diff
@@ -22,14 +22,14 @@ dependencies:
...
-    version: 2.4.0
+    version: 2.5.0
     condition: cvmfs.deploy
     alias: cvmfs
     tags:
       - deploy-cvmfs
   - name: rabbitmq-cluster-operator
     repository: https://charts.bitnami.com/bitnami
-    version: 2.6.12
+    version: 4.3.22
```
nuwang, line 32: It looks like there have been two major version changes. Have these changed the yaml configs in anyway?

afgane: I don't think so. These are the only notes I was able to find re. the release/upgrade process and nothing there seems to conflict with our setup: https://github.com/bitnami/charts/tree/main/bitnami/rabbitmq-cluster-operator#to-400

I launched 3 clusters with this version and ran a few IWC workflows, all of which worked as expected.

nuwang (approved):
Looks great! Would `outputs_to_working_directory` solve the permission issue that was encountered?

ksuderman:
> Looks great! Would `outputs_to_working_directory` solve the permission issue that was encountered?

Good questions!  I'll test it.

nuwang (approved):

PR #504, opened 2024-11-14
Update sanitize_allowlist

afgane:


nuwang (approved):

galaxy/values.yaml
```diff
@@ -600,31 +600,41 @@ configs:
     </workflow_schedulers>
   sanitize_allowlist.txt: |
     toolshed.g2.bx.psu.edu/repos/bgruening/diff/diff/3.7+galaxy0
+    toolshed.g2.bx.psu.edu/repos/bgruening/pharmcat/pharmcat/1.7.0+galaxy0
     toolshed.g2.bx.psu.edu/repos/bgruening/pharmcat/pharmcat/1.3.1+galaxy0
     toolshed.g2.bx.psu.edu/repos/crs4/taxonomy_krona_chart/taxonomy_krona_chart/2.7.1+galaxy0
+    toolshed.g2.bx.psu.edu/repos/devteam/fastqc/fastqc/0.74+galaxy1
```
dannon, line 606: Random thought -- we made these support a base path match, so you could probably consolidate safe, trusted tools like:

```
toolshed.g2.bx.psu.edu/repos/devteam/fastqc/
```

(or shorter, but per-tool seems like good granularity)

afgane: Nice, thanks for pointing that out. I'll do that in a bit.

PR #506, opened 2024-11-22
Refactor helm chart, splitting operator deps into separate chart

nuwang:
ToDo:
- [ ] Should we version the two charts separately or together?
- [ ] Should we bundle in the NFS chart with `galaxy-deps` for convenience? or disable the alien cache by default?


afgane:
I feel separate versioning is better suited for us. The Galaxy chart is likely to advance much more frequently than the deps one as we tend to not seek latest features in most of our deps.

Bundling the NFS chart seems like a reasonable thing to do so all the deps are captured.

nuwang:
Versioning separately is ok, but we would have to make the bundling logic much more sophisticated. We would probably have to detect which subfolder changed, and then bump versions accordingly. There's also the complication of which chart a  "minor/major" label applies to. Alternatively, we would need to maintain a separate chart. Maybe this is a separate thing that @ksuderman can help with, since he's done a lot of the packaging work.

afgane:
Hmm, now I'm not so sure separate versioning is not worth it because the packaging seems like a bit of magic as is placing things a various places, which leads to playing pacman to find it all... But if we version it together, how will the upgrade process on a deployed instance look? Will all the deps get rotated every time?

ksuderman:
Would it be better to separate the `galaxy-deps` chart into its own repository and package/deploy it separately?  That would clean up versioning and packaging issues.

nuwang:
Ok, I'll split it out into a new repository

afgane:
I'd just suggest naming the repo (and the chart) `galaxy-helm-deps` so it is more clear that it applies to the Helm chart.

nuwang:
Removed versioning labels because we want to get all breaking changes into v6 in one go.

nuwang:
Deps moved to: https://github.com/galaxyproject/galaxy-helm-deps

afgane (approved):
I tested this manually and installing the charts, running tools, and uninstalling the charts works as expected. Much more reliable than before!

PR #507, opened 2024-11-26
Switch to cloudnativepg

nuwang:
xref: https://github.com/galaxyproject/galaxy-helm-deps/pull/3


afgane (approved):
Tested it by deploying a couple of instances on the same cluster and ~, other than the prefix that I couldn't figure out,~ everything worked as expected.

PR #511, opened 2025-03-07
Add affinity to pgcluster

almahmoud:


nuwang (commented):
`operatorSpecExtra` allows all additional settings to be passed in. Just wondering whether you spotted that? We can add affinity in of course, but just depends on your needs.

almahmoud:
Just did this quickly when saw the conversation on slack about `nodeSelector`, didn't notice the other one, that's probably enough, can close this

PR #512, opened 2025-03-10
For single-node instances disable PDB

afgane:
CNPG docs recommend disabling the pod disruption budgets when number of instances is set to 1: https://cloudnative-pg.io/documentation/current/kubernetes_upgrade/#postgresql-clusters-used-for-development-or-testing

### DIFF
--- a/galaxy/templates/hapostgres/pgcluster.yaml
+++ b/galaxy/templates/hapostgres/pgcluster.yaml
@@ -8,6 +8,9 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   instances: {{ default "1" .Values.postgresql.instances }}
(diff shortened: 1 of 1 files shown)


galaxy/templates/hapostgres/pgcluster.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   instances: {{ default "1" .Values.postgresql.instances }}
+  {{- if or (not .Values.postgresql.instances) (eq (toString .Values.postgresql.instances) "1") }}
```
nuwang, line 11: If we specify .Values.postgresql.instances in the values file, could this then be simplified as follows?
```suggestion
  {{- if eq .Values.postgresql.instances 1 }}
```

ksuderman (approved):

Issue #515, opened 2025-04-24
timeoutSeconds is not used in multiple configurations


mrietveld:
In multiple yaml files in the helm chart, the 'failureThreshold' variable is used instead of the `timeoutSeconds` variable for the `timeoutSeconds on a startup, liveness or readiness probe. 

See the associated PR for a fix.

nuwang:
closed by: https://github.com/galaxyproject/galaxy-helm/pull/516

PR #516, opened 2025-04-24
Fix for using timeoutSeconds instead of failureThreshold for probe timeoutSeconds values

mrietveld:
See https://github.com/galaxyproject/galaxy-helm/issues/515


nuwang (approved):
Thanks for catching that and the fix!

PR #517, opened 2025-05-17
Add support for CNPG reuseExistingDirectory option

ksuderman:
Adds support for the `reuseExistingDirectory` option introduced in our [CNPG fork](https://github.com/ksuderman/cloudnative-pg/tree/main).  

Requires: https://github.com/galaxyproject/galaxy-helm-deps/pull/10


ksuderman:
We will not be using this.

PR #522, opened 2025-08-25
Switch to tpv-latest database and add chart-specific one

afgane:
Defaulting to the latest version of the TPV database and add one more link for another knob to turn for helm chart deployments.


afgane:
Re. `-latest` vs. `-v2` - this was driven by AnVIL. I wanted a place that could be updated without necessitating rolling out an update via Terra so if we version it here, the same issue remains and hence I'd prefer going with `-latest`. Conceptually too, once someone deploys the latest version of the chart, it seems they'd want the latest version of the tool resource definitions by default.

PR #531, opened 2025-11-16
Replace Ingress with Gateway API

ksuderman:
An initial attempt at what may be required to replace the Nginx Ingress controller with the Gateway API.  See the `GATEWAY_API_SUMMARY.md` and `GATEWAY_API_MIGRATION.md` files for a more detailed explanation of the changes.

These will not be breaking changes until we set `gateways.enabled: true` (or `ingress.enabled:false`) as the default.  The current default is `gateway.enabled: false`


nuwang (commented):
I'm thinking we should remove deployment specific manifests like envoy and kong - it's too many things to take maintenance on for. We are never going to even test those most likely. If we need to have an "out-of-the-box" gateway, I think istio would be a good choice, but ideally, we just keep deployment details out altogether, and only include the Gateway definition and HTTPRoute manifests (although I'm guessing that's going to end up in an ingress like scenario and we will in fact, have to rely on some preferred gateway config)

gkr0110:
I’m really excited about the possibility of using Galaxy with GatewayAPI! It would be wonderful to see this merged soon.

nuwang:
@gkr0110 Can you let us know which Gateway you are currently using?

micafer:
did you consider to use the nginx o traefik gateway api controllers?

nuwang:
@micafer I don't recall discussing those options, so thanks for bringing them up. Given the plethora of options, it just reinforces the idea that it would make more sense for the chart to be neutral about it. We could allow for multiple gateways options to be installed through the galaxy-helm-deps chart.

@afgane @ksuderman For production however, there'll be some issues like defining request timeouts, which will be gateway specific. Maybe we can just provide some notes and example config for production use cases?

For testing however, it looks like we won't need any gateway specific objects, and just an HTTPRoute alone may suffice on the galaxy-helm side: https://kubernetes.io/blog/2025/11/06/gateway-api-v1-4/#introducing-default-gateways

almahmoud:
Any plans to make changes in the k8s runner in galaxy proper for the IT ingresses as well? Assuming that will also need to change?

ksuderman:
Thanks @almahmoud.  I've opened a draft PR in Galaxy to migrate to `HTTPRoute` in the `k8s` job runner.

ksuderman:
I've removed vendor specific manifests, updated the documentation, and provided an `examples` directory with examples for Cilium, Istio, Kong, NGinx Fabrix, and Traefik.  The chart still needs testing though.

nuwang (approved):
Thanks @ksuderman, this lgtm!

galaxy/values.yaml
```diff
@@ -392,8 +392,69 @@ cronJobs:
...
+  #- timeouts and body size limits. The chart intentionally does not ship those;
+  #- see docs/gateway-api.md for example snippets you can apply alongside this chart.
+  istio:
+    timeout: "600s"
+    maxRequestHeadersKb: 96
+    streamIdleTimeout: "600s"
+    retries: {}
+      # attempts: 3
+      # perTryTimeout: 2s
+      # retryOn: 5xx,reset,connect-failure,refused-stream
```
nuwang: This section is probably a left over from earlier?
```suggestion
```

nuwang (changes requested):
I think it would be better to introduce a single docs folder and move the examples folder under it. Also left a comment on the values.yaml file.

galaxy/values.yaml
nuwang: Also @ksuderman, we are ok with the minor revision designation right? This will be backward compatible?

ksuderman: Yes, I believe this is just a `minor` upgrade as there are no breaking changes (I will double check that, and fix any breakages) and using `Ingress` is still an option.  I'll need to double check that if neither `ingress.enabled` nor `gateway.enabled` is defined then using `Ingress` is the default behaviour.

PR #533, opened 2026-02-05
Fix the upgrade process

afgane:
Two issues being addressed:
- On an upgrade, old mako templates may be cached leading to an exception during web handler startup
- With CNPG, for an unclear reason to me, it takes a little while before the database is available. On an upgrade, this causes the db upgrade script to fail. Giving it a bit of time fixes it. Not a great solution but the same flow works for both, a fresh install and an upgrade.


nuwang (approved):
LGTM!

PR #536, opened 2026-03-25
Remove stale tools and fix config typos

afgane:

PR #539, opened 2026-04-21
Helm postinstall simplified

ksuderman:
Part of the #536 replacement

A simplified version of the post install job that uses the new `abm config bootstrap` command available in version 2.12.0 of gxabm and later.

~~This won't work until the v2.12.0 of abm is pushed to quay.io~~ 
The Docker image for ABM v2.12.0 is now available on Quay.io.


galaxy/templates/rbac-job.yaml
```diff
@@ -14,6 +14,9 @@ rules:
 - apiGroups: ["batch", "extensions"]
   resources: ["jobs"]
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+- apiGroups: ["apps"]
+  resources: ["deployments"]
+  verbs: ["get", "list", "watch"]
```
nuwang: This is the wrong place for this I think. This role applies to untrusted jobs, so we should have a separate rbac role for trusted jobs.

afgane (approved):
Tested this and the current implementation works as expected. 

One idea that's tangentially related to this is whether we should change the bootstrap config structure in ABM to accommodate different sources of data than plain URL string?

For example, in addition to supporting URLs we should support imports from AnVIL/Terra workspaces. Also, parameterize a dataset with a datatype.
```
bootstrapConfig: |
    datasets:
      "RNA-seq Data":
          - "https://zenodo.org/.../SRR891268_1.fastq.gz"
          - url: "https://zenodo.org/.../annotation.gtf.gz"
            datatype: gtf
          - terra_workspace:
              namespace: "anvil-datastorage"
              name:      "AnVIL_HPRC"
              path:      "data/sample.cram"
            datatype: cram
```

Not merging this yet just because Keith wants to update how the packaging of the chart is triggered.

PR #541, opened 2026-05-12
Fix case of connection-proxy-header annotation

### LINKED ISSUE #513: Error in tusd ingress
In case of using the helm chart using new version of ingress-nginx, this error appears when installing the chart:

Error: INSTALLATION FAILED: admission webhook "validate.nginx.ingress.kubernetes.io" denied the request: annotation nginx.ingress.kubernetes.io/connection-proxy-header contains invalid value

It seems that the "Upgrade" value is not longer valid for the "nginx.ingress.kubernetes.io/connection-proxy-header" annotation:

https://github.com/galaxyproject/galaxy-helm/blob/master/galaxy/values.yaml#L780

ksuderman:
The admission webhook validator has become more strict in newer `ingress-nginx` version and rejects `Upgrade` as a value for `nginx.ingress.kubernetes.io/connection-proxy-header`.

Closes #513

### DIFF
--- a/galaxy/values.yaml
+++ b/galaxy/values.yaml
@@ -879,7 +879,7 @@ tusd:
       nginx.ingress.kubernetes.io/proxy-request-buffering: "off"
       nginx.ingress.kubernetes.io/proxy-buffering: "off"
       nginx.ingress.kubernetes.io/proxy-http-version: "1.1"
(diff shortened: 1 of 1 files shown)


nuwang (approved):
Thanks @ksuderman, lgtm!

nuwang (approved):

PR #543, opened 2026-06-11
Add missing quote 

ksuderman:
There is a missing quote character that causes `postInstallJob`s to fail on multiuser instances.


afgane:
Not related to this PR but something changed so it's taking a long time for Galaxy to start. The test failed due to a timeout. Took ~15 minutes for for it to start for me yesterday. I'll investigate tomorrow.

afgane (approved):

PR #545, opened 2026-07-09
Pre-create known database dirs

afgane:
Although Galaxy should create these dirs, I've run into a race condition between the handlers. Pre-creating them resolved it and it doesn't really cost much so adding it.


nuwang (commented):
From a workaround point of view, this is ok, but perhaps a bug should be filed upstream? Otherwise, some new database directory means a new patch here etc.

ksuderman:
I am going to merge this, but I wonder if fixing the race condition in #546 will also negate the need for this fix?

PR #546, opened 2026-07-14
Fix DB init race

ksuderman:
On an initial startup, i.e. when there is no database, the `job-init.yaml` template defers database creation to the handler pods with the message: _"No Galaxy database exists; startup will create it."_ It then writes the `init_db_done` marker.  This unblocks waiting handlers and they will race to create the database schema.

With this PR when the `init_db_done` marker is written the database initialization is actually complete and the handlers will find a properly initialized/upgraded database.

Friends with #545


nuwang (approved):
This is better! I just can't recall why we couldn't do this earlier.

PR #549, opened 2026-07-30
Fixes for when running under a proxy prefix

afgane:
Friends with https://github.com/galaxyproject/galaxy/pull/23224

### DIFF
--- a/galaxy/templates/httproute-tusd.yaml
+++ b/galaxy/templates/httproute-tusd.yaml
@@ -22,8 +22,10 @@ spec:
     {{- range .paths }}
     - matches:
         - path:
-            type: Exact
-            value: {{ .path }}
+            #- PathPrefix, not Exact: tusd serves the creation POST on the base path but
+            #- every PATCH/HEAD for a given upload on `<base path>/<upload id>`.
+            type: PathPrefix
+            value: {{ tpl .path $ }}
       backendRefs:
         - name: {{ $fullName }}-tusd
           port: 1080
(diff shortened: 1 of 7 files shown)


galaxy/templates/httproute-tusd.yaml
```diff
@@ -22,8 +22,10 @@ spec:
     {{- range .paths }}
     - matches:
         - path:
-            type: Exact
-            value: {{ .path }}
+            #- PathPrefix, not Exact: tusd serves the creation POST on the base path but
+            #- every PATCH/HEAD for a given upload on `<base path>/<upload id>`.
+            type: PathPrefix
```
nuwang: I think retaining backward compatibility here might be useful, because we are changing external contract + I think we used this in some places I can't quite recall.
```suggestion
            type: {{ .type | default "Exact" }}
```

galaxy/values.yaml
```diff
@@ -885,13 +885,25 @@ tusd:
...
+      #- uploads to galaxy-web instead of tusd.
       nginx.ingress.kubernetes.io/proxy-body-size: "0"
     hosts:
       - host: ~
         paths:
-          - path: "/galaxy/api/upload/resumable_upload"
+          #- Rendered with `tpl`, so it tracks `ingress.path` automatically. It must
+          #- stay in sync with the `-base-path` tusd is started with, otherwise the
+          #- route matches nothing and uploads silently fall through to galaxy-web.
+          - path: '{{ include "galaxy.add_trailing_slash" .Values.ingress.path }}api/upload/resumable_upload'
```
nuwang, line 899:
```suggestion
          - path: '{{ include "galaxy.add_trailing_slash" .Values.ingress.path }}api/upload/resumable_upload'
             type: PathPrefix          
```

nuwang (commented):
Made 2 tweaks to path prefix handling. Otherwise lgtm.

PR #553, opened 2026-08-26
Bump Galaxy to 26.1.1

afgane:


nuwang:
The action will bump the version. I think I made a mistake during the previous merge, because I should have added the patch label. @ksuderman Is there some other way to trigger a new release?

afgane:
I think we're good: we just merge to `dev` and then when ready to cut a release, open a PR to `master` and tag that PR (https://github.com/galaxyproject/galaxy-helm/blob/dev/docs/release-process.md). So once this is merged, I was going to open that PR.

ksuderman:
> @ksuderman Is there some other way to trigger a new release?

@nuwang The release workflows only bumps the chart version.  The `appVersion` and `image.tag` still need to be set manually.

The workflow has a `workflow_dispatch` trigger so it can be run manually if needed.

PR #556, opened 2026-08-28
Link legacy .loc files from reference data into tool_data_path

afgane:
## Problem

A handful of legacy Galaxy tools never go through data tables. They take `-g '${GALAXY_DATA_INDEX_DIR}'` and open `<tool_data_path>/<name>.loc` **by path** — `GALAXY_DATA_INDEX_DIR` being `tool_data_path` (`lib/galaxy/tools/evaluation.py:650`).

The chart already wires `tool_data_table_config_path` at the reference data, so table-driven tools are fine. But that never reaches these tools, and `tool_data_path` holds only `.loc.sample` files on a fresh deployment. So they fail outright:

```
FileNotFoundError: [Errno 2] No such file or directory:
  '/galaxy/server/database/tool-data/alignseq.loc'
```

| tool | needs |
|---|---|
| Extract Genomic DNA, BLAT | `alignseq.loc`, `twobit.loc` |
| Interval2Maf, MAF stats, GeneBed/Interval MAF to FASTA | `maf_index.loc` |
| Interval2Maf pairwise | `maf_pairwise.loc` |
| Microbial data import | `microbial_data.loc` |
| SIFT | `sift_db.loc` |
| SRMA | `srma_index.loc` |

All seven ship in the reference data under `byhand/location`.

## Change

`templates/jobs-init.yaml` already mounts the Galaxy PVC and the reference data in the same pod, so this adds an init container that symlinks each file into `tool_data_path` when `refdata.enabled`.

Symlinks rather than copies: the files stay in step with the reference data and `microbial_data.loc` alone is 10k lines.

Deliberately **best effort** — it never exits non-zero, skips files the reference data does not carry, and never replaces an existing real file. A missing or operator-supplied `.loc` cannot block startup.

The list is exposed as `refdata.legacyLocFiles`; set it to `[]` to disable. Not repointing `tool_data_path` at the reference data, since it must stay writable for shed tools and data managers.

## Validation

`helm lint` and `helm template` pass. Gating verified: absent with `refdata.enabled=false`, absent with `legacyLocFiles` empty, and the CVMFS-only `subPath`/`mountPropagation` correctly omitted for `refdata.type=s3csi`.

Ran the **rendered** script on a live k8s deployment with CVMFS reference data:

- all seven linked, running as the `galaxy` user
- second run reports all seven already present (idempotent)
- replacing one with a real file leaves it untouched, content preserved
- an unknown filename is skipped, exit 0

Then `Extract Genomic DNA` on an hg38 BED, `index_source=cached` — the exact path that previously raised `FileNotFoundError`:

```
state: ok   exit: 0   output: fasta, 188 bytes

>hg38_chr1_1000000_1000060_+ f1
GTGGAGCGCGCCGCCACGGACCAcgggcgggctggcgggcgagcggcgag
cgcgcggCGA
>hg38_chr1_2000000_2000060_- f2
tcctgaggagctgggactccaggcgcctgcaccacacctggctaattttt
tgtattttta
```

Correct coordinates, soft-masking preserved, minus-strand feature reverse-complemented.

## Coverage of the referenced data

Sampled paths out of each file against the public repository. `alignseq.loc` (859 `seq` entries, hg38/hg19/hg18/mm10/mm9/sacCer3 all present), `maf_index.loc`, `maf_pairwise.loc`, `microbial_data.loc`, `sift_db.loc` and `srma_index.loc` all resolve fully.

`twobit.loc` is only partly backed — 8 of 26 sampled. It is consulted solely as a fallback when `alignseq.loc` has no entry for a genome, so linking it can only add working genomes; omitting it would strictly remove them.

## Note

This fixes the `index_source=cached` branch of Extract Genomic DNA. Its `history` branch builds a 2bit via `faToTwoBit`, which the Galaxy image does not carry, and the tool sits on `GALAXY_LIB_TOOLS_VERSIONED` below the threshold so container resolvers are skipped for it — that branch needs Galaxy's Python environment *and* an external binary, so no container choice satisfies it. Out of scope here.

### DIFF
--- a/galaxy/templates/_helpers.tpl
+++ b/galaxy/templates/_helpers.tpl
@@ -353,23 +353,3 @@ parentRefs:
     namespace: {{ .Release.Namespace }}
 {{- end }}
 {{- end -}}
-
-{{/*
-The effective tool_data_path. Galaxy's own config is the source of truth, so
-read it back from configs."galaxy.yml".galaxy.tool_data_path and fall back to
-the chart default when it is unset. That value may itself contain template
-syntax, hence the tpl.
-
-Never reference this helper from tool_data_path itself - it reads that key, so
-doing so would recurse.
-*/}}
-{{- define "galaxy.toolDataPath" -}}
-{{- $cfg := (index (.Values.configs | default dict) "galaxy.yml") | default dict -}}
-{{- $galaxy := (index $cfg "galaxy") | default dict -}}
-{{- $path := index $galaxy "tool_data_path" -}}
-{{- if $path -}}
-{{- tpl (printf "%v" $path) . -}}
-{{- else -}}
-{{- printf "%s/tool-data" .Values.persistence.mountPath -}}
-{{- end -}}
-{{- end -}}
--- a/galaxy/templates/_helpers.tpl
+++ b/galaxy/templates/_helpers.tpl
@@ -353,3 +353,23 @@ parentRefs:
     namespace: {{ .Release.Namespace }}
 {{- end }}
 {{- end -}}
+
+{{/*
+The effective tool_data_path. Galaxy's own config is the source of truth, so
+read it back from configs."galaxy.yml".galaxy.tool_data_path and fall back to
+the chart default when it is unset. That value may itself contain template
+syntax, hence the tpl.
+
+Never reference this helper from tool_data_path itself - it reads that key, so
+doing so would recurse.
+*/}}
+{{- define "galaxy.toolDataPath" -}}
+{{- $cfg := (index (.Values.configs | default dict) "galaxy.yml") | default dict -}}
+{{- $galaxy := (index $cfg "galaxy") | default dict -}}
+{{- $path := index $galaxy "tool_data_path" -}}
+{{- if $path -}}
+{{- tpl (printf "%v" $path) . -}}
+{{- else -}}
+{{- printf "%s/tool-data" .Values.persistence.mountPath -}}
+{{- end -}}
+{{- end -}}
--- a/galaxy/templates/jobs-init.yaml
+++ b/galaxy/templates/jobs-init.yaml
@@ -72,25 +72,38 @@ spec:
               # Best effort on purpose: a missing .loc must not block startup,
               # so this never exits non-zero and never replaces a real file.
               echo "[`date`][init-tool-data] - Linking legacy .loc files into tool_data_path. Running as `id`."
-              TOOL_DATA="{{ .Values.persistence.mountPath }}/tool-data"
+              TOOL_DATA="{{ include "galaxy.toolDataPath" . }}"
               mkdir -p "$TOOL_DATA"
               for f in {{ .Values.refdata.legacyLocFiles | join " " }}; do
-                if [ -e "$TOOL_DATA/$f" ]; then
+                # -e follows the link, so a dangling symlink reads as absent
+                # here but still blocks ln. Test -L separately and replace only
+                # those, leaving real files and working links untouched.
+                if [ -L "$TOOL_DATA/$f" ] && [ ! -e "$TOOL_DATA/$f" ]; then
+                  echo "[`date`][init-tool-data] - $f is a dangling symlink, replacing it."
+                  rm -f "$TOOL_DATA/$f"
+                elif [ -e "$TOOL_DATA/$f" ]; then
                   echo "[`date`][init-tool-data] - $f already present, leaving it as is."
                   continue
                 fi
+                src=""
                 for d in /cvmfs/data.galaxyproject.org/byhand/location /cvmfs/data.galaxyproject.org/managed/location; do
                   if [ -f "$d/$f" ]; then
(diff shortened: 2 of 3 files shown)


afgane:
Thanks — both confirmed by reproduction, both fixed in 56010c3.

## Dangling symlinks

Reproduced exactly as described. `[ -e ]` follows the link, so a dangling one reads as absent and the skip branch is not taken; the path still exists, so `ln -s` fails:

```
ln: .../tool-data/alignseq.loc: File exists
  alignseq.loc not found in reference data, skipping.
after: still dangling=YES
```

Note the second line was also wrong — the file *was* in the reference data. The `ln` failure and a genuinely missing source shared one message, so the log actively pointed away from the real cause.

Now `-L` is tested separately, so only dangling links are removed and recreated; real files and working links are still left alone. I also split the source lookup out of the `ln` so those two cases report distinctly:

```
if [ -L "$TOOL_DATA/$f" ] && [ ! -e "$TOOL_DATA/$f" ]; then
  echo "... $f is a dangling symlink, replacing it."
  rm -f "$TOOL_DATA/$f"
elif [ -e "$TOOL_DATA/$f" ]; then
  ... continue
fi
```

## Configured tool_data_path

Also confirmed. Rendering with `configs."galaxy.yml".galaxy.tool_data_path=/custom/tool-data` gave:

```
tool_data_path: /custom/tool-data          <- Galaxy config
TOOL_DATA="/galaxy/server/database/tool-data"   <- init container
```

Added a `galaxy.toolDataPath` helper that reads the value back from Galaxy's own config and falls back to the chart default, used for both the script and the volume mount.

**One thing I went slightly beyond the report on**, so flagging it for you to push back on if you disagree: the job runner's PVC spec in `job_conf` (`values.yaml:592`) hardcoded the same path. Left alone, job pods would still mount `tool-data` at the default path and never see the links, so honouring the config in the init container alone would have been half a fix. It now uses the helper as well. Default rendering is byte-identical; with the override, all four agree:

```
tool_data_path: /custom/tool-data                 <- Galaxy config
TOOL_DATA="/custom/tool-data"                     <- init script
mountPath: /custom/tool-data                      <- init volume mount
...pvc/tool-data:/custom/tool-data:rw             <- job runner PVC spec
```

The helper reads `tool_data_path`, so that value must not reference the helper back — called out in a comment on the define to prevent recursion.

## Re-validation

Full state matrix against the rendered script: dangling link replaced, valid link and real file untouched, a file present only in `managed/location` picked up, absent source skipped, second run creates nothing, exit 0 throughout.

On a live deployment with real CVMFS, a deliberately dangling `alignseq.loc`:

```
before: alignseq.loc dangling=YES
  - alignseq.loc is a dangling symlink, replacing it.
  - linked alignseq.loc -> /cvmfs/data.galaxyproject.org/byhand/location/alignseq.loc
after: resolves=YES
```

`Extract Genomic DNA` on an hg38 BED still returns exit 0 afterwards. `helm lint` and `helm template` pass.

afgane:
Agreed — and confirmed. Rendering with `tool_data_path=/custom/tool-data` and checking every workload for a mount at that path:

```
HAS      test-galaxy-init-db-bgtit     ['galaxy-init-tool-data']
MISSING  test-galaxy-web               []
MISSING  test-galaxy-job-0             []
MISSING  test-galaxy-workflow          []
MISSING  test-galaxy-celery            []
MISSING  test-galaxy-celery-beat       []
```

All five deployments mount only the whole PVC at `persistence.mountPath` plus a `config` subPath. So my previous commit made a custom path work in two places out of seven — and the job handler, where the local runner executes, was one of the misses.

**Taken the "custom paths remain unsupported" option**, reverted in 1a60aeb.

Reasoning: making it work properly means adding a subPath mount at the effective path to all five long-running deployments. That is a chart-wide feature rather than part of linking `.loc` files, and — more to the point — it is the one part of this change I could not validate to the same standard as the rest. Everything else here was exercised on a live cluster; a five-deployment mount change I would only have been able to render, not run. Shipping that unvalidated into every long-running pod is a poor trade for a config knob nobody in this PR needs.

So the previous commit's helper and both call sites are reverted. Before this PR a custom `tool_data_path` was uniformly ignored; it is uniformly ignored again, rather than working inconsistently. The assumption is now written down in the values docs instead of being implicit:

```
#- Like the rest of the chart, this assumes tool_data_path is left at its
#- default of `{persistence.mountPath}/tool-data`. Overriding
#- `configs."galaxy.yml".galaxy.tool_data_path` is not supported: the
#- long-running Galaxy pods mount only the whole PVC at
#- `persistence.mountPath`, so a custom path would not exist inside them.
```

Happy to do proper custom-path support as a separate PR if it is wanted — it would need the subPath mount in all five deployments plus the job runner PVC spec, and ideally a deploy that actually exercises a non-default path.

The dangling-symlink fix is unaffected and still verified after the revert: dangling replaced, valid link and real file untouched, `managed/location`-only file picked up, absent source skipped, second run a no-op, exit 0 throughout. Default rendering is unchanged; `helm lint` and `helm template` pass.

PR #563, opened 2026-09-10
Put Galaxy's lib on PYTHONPATH for built-in tools pinned to the image

afgane:
## The problem

The `force_default_container_for_built_in_tools` TPV rule pins built-in tools to the Galaxy image, on the premise stated in its own comment that *"the Galaxy tree under /galaxy/server/lib exists solely inside this image"*.

The tree is there — but it is never installed into the venv, and job containers inherit no `PYTHONPATH`. So `import galaxy` fails even though `python` correctly resolves to the venv:

```
$ python -c "import galaxy"
ModuleNotFoundError: No module named 'galaxy'

$ PYTHONPATH=/galaxy/server/lib python -c "from galaxy.util.compression_utils import CompressedFile"
(ok)
```

Both probed inside `galaxy-min:26.1-auto`, where `PATH` already begins with `/galaxy/server/.venv/bin`.

`CONVERTER_archive_to_directory` is the case that surfaced it. Its command is:

```
python -c "from galaxy.util.compression_utils import CompressedFile; CompressedFile('$input1').extract('.')"
```

which dies with `ModuleNotFoundError`, exit 1, leaving the converted dataset in `error`. That breaks unpacking any archive to a directory datatype — `zarr`, `ome_zarr`, `directory` — on a Kubernetes deployment.

Every tool the rule matches has the same exposure, since needing that tree is precisely what puts it there: `CONVERTER_tar_to_directory` runs the identical command, and `GALAXY_LIB_TOOLS_UNVERSIONED` entries are selected for exactly this reason.

## The change

Add `env: PYTHONPATH: /galaxy/server/lib` to that rule. It is scoped to the rule, so tools resolved to a biocontainer are unaffected.

## Testing

On a Kubernetes deployment, `CONVERTER_archive_to_directory` goes from failing immediately to `ok`, and a `zarr` dataset reaches `ok` with `store_root` and `zarr_format` metadata correctly set — verified end-to-end by running the `ip_binary_to_labelimage` tool tests, which need a zarr input, against that deployment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dannon:
Should we have just done this in the Dockerfile in the first place? I had added PYTHONPATH in https://github.com/galaxyproject/galaxy/pull/21593, but we closed that with this part unresolved. Seems like the same issue?

ksuderman:
[**Edited to add**] Claude is getting a little ahead of itself and posted this *on my behalf* without my approval (I have sternly chastised Claude and it has promised to never do that again... 🙄 Having said that, I think Claude may have a good point.

Same root cause, yes. The image has never set `PYTHONPATH`, and the chart has been compensating for that on the handler side all along: every handler pod gets `PYTHONPATH=/galaxy/server/lib` from `galaxy.podEnvVars` in `_helpers.tpl`, and the web handler additionally passes `--pythonpath /galaxy/server/lib` to gunicorn. Job pods were the one place we were not doing it, because their environment comes from the TPV destination rather than from the handler pod.

I think the chart is the right place for this fix regardless of what happens in the Dockerfile, for two reasons:

1. Scope. The unresolved concern on galaxyproject/galaxy#21593 was whether an image-wide `PYTHONPATH` messes with the tool execution environment. Putting it on this TPV rule sidesteps that: it only reaches tools that are pinned to the Galaxy image precisely because they need the Galaxy lib. Nothing resolved to a biocontainer, and nothing a deployer pins to the image for some other reason, sees it.

2. Coverage. It works with the image tags people are already running without waiting for a rebuild, and it ships in a chart release rather than a Galaxy release.

Adding `ENV PYTHONPATH` to `.k8s_ci.Dockerfile` as well would be fine by me and would help anyone using the image outside the chart. If that lands, this rule just becomes redundant with the same value, so the two do not conflict. I would rather not have this fix wait on reopening that discussion upstream though.

ksuderman (approved):

afgane:
I'm in favor of merging this because it's contained to the listed converters and quick to deploy. However, I also don't see why not put it in the base Dockerfile. If a tool using this image needs a python env, it'll be Galaxy's - there is no other python env there to use. I don't know if any tools create their own envs at runtime, which seems like the only place setting PYTHONPATH in the base image could go off rails, no?

dannon:
Yeah, for clarity I think it's totally fine to merge this -- it just reminded me that I had the same problem and had tried to fix it in the dockerfile before.  I don't know if there's a real use case where someone would *not* want that set, but maybe - kind of validating as a good piece of evidence for the other fix to see that others have the same issue, though.

PR #570, opened 2026-09-18
Only pre-create legacy database/files dir when it already exists

afgane:
## Summary

- PR #545 added an `mkdir -p /galaxy/server/database/files` in the init
  container to fix a race between concurrent handler pods each trying to
  create the dataset directory independently.
- Hardcoding `database/files` there means every fresh install is forced
  onto the pre-20.05 Galaxy default. Galaxy core changed the dataset dir
  default from `files` to `objects` in v20.05, and picks `store_by=uuid`
  instead of `store_by=id` specifically when the directory is named
  `objects` - so fresh Helm installs have been silently missing that
  UUID-based storage scheme.
- Galaxy core's own fallback (`_set_alt_paths`) decides which name to use
  by checking whether `database/files` already exists on disk, so this PR
  makes the init container do the same check before creating the
  directory: use `database/files` if it's already there, otherwise create
  `database/objects`.
- Existing deployments are unaffected (their `database/files` dir already
  exists, so the check keeps them on the same path). Only genuinely fresh
  installs change, picking up Galaxy's current default. The single
  init-container pre-creation that #545 introduced is preserved, so the
  original race is still fixed.

## Test plan

- [ ] `helm lint galaxy` passes
- [ ] `helm template galaxy galaxy` renders the expected conditional in
      the init container script
- [ ] Deploy against a fresh PVC and confirm Galaxy comes up using
      `database/objects` with `object_store_store_by: uuid`
- [ ] Deploy against an existing PVC with `database/files` already
      populated and confirm Galaxy still finds its datasets there

<sub>Summary compiled by Claude (Claude Code), working with @afgane.</sub>


galaxy/templates/_helpers.tpl
```diff
@@ -198,8 +198,9 @@ mkdir -p /galaxy/server/database/cache/compiled_templates;
 mkdir -p /galaxy/server/database/tmp;
 mkdir -p /galaxy/server/database/object_store_cache;
 mkdir -p /galaxy/server/database/jobs_directory;
-mkdir -p /galaxy/server/database/files;
-chown -R {{ .Values.securityContext.runAsUser }}:{{ .Values.securityContext.runAsUser }} /galaxy/server/database/cache /galaxy/server/database/tmp /galaxy/server/database/object_store_cache /galaxy/server/database/jobs_directory /galaxy/server/database/files;
+if [ -d /galaxy/server/database/files ]; then GALAXY_DATASET_DIR=/galaxy/server/database/files; else GALAXY_DATASET_DIR=/galaxy/server/database/objects; fi;
+mkdir -p "$GALAXY_DATASET_DIR";
+chown -R {{ .Values.securityContext.runAsUser }}:{{ .Values.securityContext.runAsUser }} /galaxy/server/database/cache /galaxy/server/database/tmp /galaxy/server/database/object_store_cache /galaxy/server/database/jobs_directory "$GALAXY_DATASET_DIR";
```
nuwang, line 203:
```suggestion
chown -R {{ .Values.securityContext.runAsUser }}:{{ .Values.securityContext.runAsGroup }} /galaxy/server/database/cache /galaxy/server/database/tmp /galaxy/server/database/object_store_cache /galaxy/server/database/jobs_directory "$GALAXY_DATASET_DIR";
```

nuwang (approved):

